Issue #39, opened 2014-04-17
Need relaxing option for empty alt check


nschonni:
Alt tags may be left empty for decorative images http://dev.w3.org/html5/alt-techniques/#secm3
It would be good if to have a similar whitelist as the href check to exclude known valid images like "logo.png" where the content is purely decorative. In the decorative case, an empty alt tag prevents screen readers from reading the file name as a fallback.

LaurentGoderre:
actually, it should be `alt=""` because `alt=" "` is a non null alt and will not be ignored by assistive technologies

PR #62, opened 2014-07-20
Ignore missing alt tags when asked

gjtorikian:
Closes https://github.com/gjtorikian/html-proofer/issues/39


LaurentGoderre:
If I read this correctly, the tests for this are incorrect

This should be an error

```html
<img src="..."/>
```

This shouldn't be

```html
<img src="..." alt=""/>
```

It is an error to not have an alt at all it's not an error to have a null one.

gjtorikian:
I could see `alt=""` being flagged as an error, sure. An empty length `alt` tag is the same as a missing `alt` tag. I'll log an issue to fix.

PR #137, opened 2015-01-04
Support for non-UTF-8 encoded URLs

benbalter:
This fixes a conversion error that gets raised when running HTML-proofer on non-UTF-8 encoded URLs.


gjtorikian:
Is it possible to add a test? If not, just curious to know which file/URL was not working.

benbalter:
Yes. Will do. The URL that tripped it was http://ben.balter.com/2014/09/29/source-disclosed-%E2%89%A0-open-source/.

gjtorikian:
After spending too long trying to write a test for the two cases (https://github.com/gjtorikian/html-proofer/blob/38a2d9c519d01d2fed142961fa618d72d30492bd/spec/html/proofer/fixtures/links/utf8Link.html), I finally got it. I love `Addressable`.

benbalter:
Nice. Thanks for tackling that. Yeah, we use Addressable as part of Gman. Works wonders.

PR #142, opened 2015-01-11
Don't use deprecated method

myokoym:
The following warning occurred in the script in README.md:

    % ruby -v -W -e 'Dir.mkdir("out") unless File.exists?("out")'
    ruby 2.1.5p273 (2014-11-13 revision 48405) [x86_64-linux]
    -e:1: warning: File.exists? is a deprecated name, use File.exist? instead

I tried not to use a deprecated method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ require 'html/pipeline'
 require 'find'
 
 # make an out dir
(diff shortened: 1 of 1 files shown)


penibelst:
Thank you.

Issue #131, opened 2014-11-11
Add throttle options


mojavelinux:
When running html-proofer on a site, I'm bumping into what appears to be rate limits from the URL endpoints being checked (most notably, raw.githubusercontent.com).As a result, I'm seeing a lot of false broken link errors (error codes 400 and 0).

To address this problem, I think there needs to be some way to control how many concurrent connections are used and how much time is waited between requests.

One way seems to be to allow the `max_concurrency` flag on Hydra to be set. If I hack the html-proofer gem, I can minimize the false errors by reducing the number of concurrent requests:

```
@hydra ||= Typhoeus::Hydra.new(max_concurrency: 1)
```

I still hit the rate limits eventually, but that allows many more links to pass. We also need to be able to set a wait time in between checks against the same domain name.

mojavelinux:
:+1:

Issue #129, opened 2014-11-04
Refactor `run` method


gjtorikian:
The `run` method is crazy long. Refactor it into smaller chunks as suggested here: https://github.com/gjtorikian/html-proofer/pull/116#issuecomment-55528921

Issue #114, opened 2014-09-12
Namespace Typhoeus options


gjtorikian:
As noted in https://github.com/gjtorikian/html-proofer/pull/113#issuecomment-55334811, Typhoues is real picky about what it take in. I'll need to make a breaking release in order to namespace Typhoeus (and other libs!) options. So rather than

```
HTML::Proofer.new("out/", {:ext => ".htm", :verbose => true, :ssl_verifyhost => 2 })
```

It should be

```
HTML::Proofer.new("out/", {:ext => ".htm", :verbose => true, :typhoeus => {:ssl_verifyhost => 2 }})
```

holic:
:+1:

PR #147, opened 2015-01-24
Refactor html-proofer

gjtorikian:
This refactors html-proofer into much more manageable and smaller chunks. Notably, it:

* Moves the logging logic into its own class.
* Moves the external URL checking logic into its own class.
* Renames the horrible `HTML::Proofer::Checks::Check` to a more sane `HTML::Proofer::Runner`.
* Simplifies custom test logic

It also namespaces Typheous/Hydra option passing; hence this will be a major version bump when merged.

Fixes https://github.com/gjtorikian/html-proofer/issues/131, fixes https://github.com/gjtorikian/html-proofer/issues/114, fixes https://github.com/gjtorikian/html-proofer/issues/129.

Although none of you are under any obligation whatsoever to do so, a review from @penibelst / @benbalter / @parkr might be nice, to make sure I'm not missing something.


bin/htmlproof
```diff
@@ -29,31 +29,31 @@ Mercenary.program(:htmlproof) do |p|
...
+    options[:href_ignore] = opts['href_ignore'] unless opts['href_ignore'].nil?
+    options[:alt_ignore] = opts['alt_ignore'] unless opts['alt_ignore'].nil?
+    options[:file_ignore] = opts['file_ignore'] unless opts['file_ignore'].nil?
+    options[:disable_external] = opts['disable_external'] unless opts['disable_external'].nil?
+    options[:only_4xx] = opts['only_4xx'] unless opts['only_4xx'].nil?
+    options[:favicon] = opts['favicon'] unless opts['favicon'].nil?
+    options[:verbose] = opts['verbose'] unless opts['verbose'].nil?
+    options[:directory_index_file] = opts['directory_index_file'] unless opts['directory_index_file'].nil?
+    options[:validate_html] = opts['validate_html'] unless opts['validate_html'].nil?
+    options[:check_external_hash] = opts['check_external_hash'] unless opts['check_external_hash'].nil?
```
parkr: can this be DRY'd?

parkr:
:heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart:

bin/htmlproof
gjtorikian: ugh probably. I was thinking about removing the bin entirely but I guess that's Not The Right Thing To Do :tm:.

penibelst:
Really nice.

With smaller chunks we could sometimes test classes instead of the whole thing. I’ve added tests for the new Utils class in 700613208843c72b6d06b6400fb8616186040183.

penibelst:
Nice run.

PR #148, opened 2015-01-24
Create a Rake task

penibelst:
@gjtorikian According to your comment

> I was thinking about removing the bin entirely

What do you think about introducing a Rake task? I created a simple example inspired by [Flay](https://github.com/seattlerb/flay/blob/master/lib/flay_task.rb). It’s not tested.

### DIFF
--- a/lib/html/proofer/task.rb
+++ b/lib/html/proofer/task.rb
@@ -1,0 +1,23 @@
+class HTML::Proofer::Task < Rake::TaskLib
+
+  attr_accessor :name, :src, :verbose
+
+  def initialize name = :proofer, src
+    @name = name
+    @src  = src
+    @verbose = Rake.application.options.trace
+
+    yield self if block_given?
+
+    define
+  end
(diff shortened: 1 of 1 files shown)


parkr:
Usage?

gjtorikian:
I am not sure I am in favor of adding more entry points, not less. Using the program already within a Rake task is pretty simple to do.

My only interest in maintaining a bin is so that it can be used outside of Ruby projects--within any project that generates HTML, really. Introducing a Rake task locks it down once more towards Ruby-only usage.

penibelst:
@parkr The usage in a rakefile:

```ruby

BUILD_DIR = "build"
directory BUILD_DIR
file BUILD_DIR do
  Foo::Bar::generate BUILD_DIR
end

HTML::Proofer::Task :proofer => ["build"] do |t|
  t.src = BUILD_DIR
end
```

@gjtorikian I understand your point.

Issue #146, opened 2015-01-20
No line numbers in error output


mietek:
When using `htmlproof` from the command-line, the error output for my sites consists mostly of the following messages:

```
  *  anchor has no href attribute
  *  anchor has no href attribute
  *  anchor has no href attribute
...
```

Please add line numbers to error output in order to assist with fixing these errors.

gjtorikian:
I am pretty sure this is impossible to do, as html-proofer is using Nokogiri, and Nokogiri just looks at nodes, not content.

mietek:
It seems like Nokogiri can, in fact, keep track of the line numbers while parsing:

> ### [Nokogiri::XML::Node#line](http://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Node:line)
> 
> ```
> #line ⇒ Object
> ```
> Returns the line for this Node

gjtorikian:
Huh. I'll look into it.

mietek:
Thanks.

PR #154, opened 2015-01-29
Add basic line number support

gjtorikian:
Should close https://github.com/gjtorikian/html-proofer/issues/146 when it's ready.


gjtorikian:
Note that at this time, external links won't be reported, because they are pulled out and processed separately, for performance reasons.

Issue #158, opened 2015-01-29
Support HTML5 data attributes


mietek:
Currently, proofing a project which uses [HTML5 data attributes](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes) results in errors:
```
htmlproof 2.0.1 | Error:  `@data-target' is not allowed as an instance variable name
```
**Note:** This worked in an earlier version of `html-proofer`.

mietek:
Thanks.

PR #159, opened 2015-01-29
Translate attributes with dashes into underscores

gjtorikian:
Closes https://github.com/gjtorikian/html-proofer/issues/158.

@mietek You keep logging them, I'll keep fixing them. :wink:


mietek:
I’m afraid that’s all I got. Thanks!

PR #167, opened 2015-02-03
Add missing cgi dependency

johnelse:
`htmlproof --check-html` tries to call CGI.unescape, which is not in scope
at the moment.


gjtorikian:
Ugh, thanks!

penibelst:
@gjtorikian I see now the `Addressable::URI.unencode` method. Should we use it rather and avoid the CGI dependency?

PR #178, opened 2015-02-19
perform url substitutions in src attributes also

kyleboyle:
Static site generators have the option of generating urls with a baked-in base url. When using this type of option, links have absolute urls which will not resolve when checking locally. When performing checks against a static site on local filesystem the href_swap option can be used to properly check these links. Example content:
```html
...
<link href="http://baseurl.com/index.xml" rel="alternate" type="application/rss+xml" title="Website Title RSS" />
...
<a href="http://baseurl.com/license">lic</a>
...
<script src="http://baseurl.com/js/jquery.min.js"></script>
...
```

Link check will fail with vanilla htmlproof config, but using the following config will work for the hrefs:
```{:href_swap => {/http:\/\/baseurl\.com/ => ""} ```

The script tag will still fail however.  This pr adds the same regex substition to the src attributes.

### DIFF
--- a/spec/html/proofer/fixtures/images/replaceAbsUrlSrc.html
+++ b/spec/html/proofer/fixtures/images/replaceAbsUrlSrc.html
@@ -1,0 +1,2 @@
+
+<img alt="sure thing" src="http://baseurl.com/gpl.png"/> base url is relative to root
--- a/spec/html/proofer/fixtures/images/replaceAbsUrlSrc.html
+++ b/spec/html/proofer/fixtures/images/replaceAbsUrlSrc.html
@@ -1,2 +1,2 @@
 
-<img alt="sure thing" src="http://baseurl.com/gpl.png"/> base url is relative to root
+<img alt="sure thing" src="http://baseurl.com/gpl.png"/>
--- a/lib/html/proofer/checkable.rb
+++ b/lib/html/proofer/checkable.rb
@@ -22,7 +22,11 @@ module HTML
         if @href && @check.options[:href_swap]
           @href = swap(@href, @check.options[:href_swap])
         end
-
+        
+        if @src && @check.options[:href_swap]
+          @src = swap(@src, @check.options[:href_swap])
+        end
+        
         # fix up missing protocols
         @href.insert 0, 'http:' if @href =~ %r{^//}
         @src.insert 0, 'http:' if @src =~ %r{^//}
--- a/spec/html/proofer/fixtures/scripts/script_abs_url.html
+++ b/spec/html/proofer/fixtures/scripts/script_abs_url.html
@@ -1,0 +1,9 @@
+<html>
+
+<body>
+
+  <script src="http://baseurl.com/script.js"></script>
+
+</body>
+
+</html>
--- a/spec/html/proofer/fixtures/scripts/script_abs_url.html
+++ b/spec/html/proofer/fixtures/scripts/script_abs_url.html
@@ -2,7 +2,7 @@
 
 <body>
 
-  <script src="http://baseurl.com/script.js"></script>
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by gjtorikian** [comment]
Interesting. Can you add a test that implements this?

**Comment by kyleboyle** [comment]
Probably not, I've never used ruby before

**Comment by kyleboyle** [comment]
@gjtorikian added src tests for script and image

File: spec/html/proofer/fixtures/images/replaceAbsUrlSrc.html [hunk 1]
```diff
@@ -0,0 +1,2 @@
+
+<img alt="sure thing" src="http://baseurl.com/gpl.png"/> base url is relative to root
```
penibelst: Please replace `http://baseurl.com` with `http://example.com` everywhere. You can omit the comment “ base url is relative to root” too.

penibelst:
> added src tests for script and image

Great.

mdotasia:
This seems to be a problem for images and scripts.

SamuelDebruyn:
What is the recommended way to go if I want to run html-proofer on the source of my static website? With the default settings, I get errors when I'm publishing a new post. (Pages that reference that new post contain *dead* links because the post isn't live yet.)

parkr:
> (Pages that reference that new post contain dead links because the post isn't live yet.)

How do pages reference the post if the post isn't published?

SamuelDebruyn:
I should have mentioned that I'm talking about static websites generated with Jekyll, Hugo, Octopress or something similar.

The website/blog [I'm talking about](https://github.com/SamuelDebruyn/sa.muel.be-hugo) is made with [hugo](http://github.com/spf13/hugo).

So I write my theme and all the posts and then generate a static website. The theme uses `.Site.BaseURL` to refer to the baseURL of the website which can be set in the config file. When building, hugo replaces those URLs with absolute URLs. The problem happens when I add content to the website. E.g. my `.Site.baseURL` is set to `http://sa.muel.be` so the generated HTML files will contains links to `http://sa.muel.be/my-new-post`. However, that URL won't work until I publish the newly generated static website. If I'd then run html-proofer on the generated website, it would tell me that `http://sa.muel.be/my-new-post` returns 404.

I'd like to use html-proofer as one of my build steps, but this way, I can never add new content to my website.

kyleboyle:
Use href_swap to replace your base url with nothing, making your links relative. The point of this pull request is to extend that swapping functionality to other places that may use the base url.

SamuelDebruyn:
Could you give a short example of the command line parameter I need? I tried this a couple of times but haven't succeeded so far. Thanks in advance!

kyleboyle:
I don't believe there is a cmd line that will work because it uses a colon to separate the match and replace, but our match contains a colon. you'll just have to write a node script that invokes htmlproof with the href_swap config.

SamuelDebruyn:
Would this be a useful feature for your Wercker step? Because it seems like I can't use that one right now...

kyleboyle:
there are 2 problems with the cmd line that you can work around.
1. can't use colons in regex
2. can't use a blank string for the replacement

this should work for you
    htmlproof public --href-swap "http.\/\/localhost.1313/:/"
this will replace `http://localhost:1313/` with `/` in hrefs

I'd hate to implement workarounds for htmlproofer in the wercker step, it'd probably be best to fix the issues in htmlproofer.

gjtorikian:
Thanks for these tests! I'm going to close this down because I'm working on the next version of HTML-Proofer, and it's easier to point changes to that branch.

I've cherry picked your commits and made some appropriate changes so that you'll retain contribution credit: https://github.com/gjtorikian/html-proofer/pull/297

Thanks again!

Issue #173, opened 2015-02-13
Crash with "`@:' is not allowed as an instance variable name"


tkrotoff:
Using v2.1.0 with a Jekyll website and the Rake snippet from the documentation. (gems installed inside vendor/bundle).

```
Running ["ImageCheck", "LinkCheck", "ScriptCheck"] checks on ./_site on *.html... 


rake aborted!
`@:' is not allowed as an instance variable name
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer/checkable.rb:13:in `instance_variable_set'
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer/checkable.rb:13:in `block in initialize'
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer/checkable.rb:12:in `each_pair'
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer/checkable.rb:12:in `initialize'
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer/checks/links.rb:31:in `new'
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer/checks/links.rb:31:in `block in run'
/vendor/bundle/ruby/2.0.0/gems/nokogiri-1.6.6.2/lib/nokogiri/xml/node_set.rb:187:in `block in each'
/vendor/bundle/ruby/2.0.0/gems/nokogiri-1.6.6.2/lib/nokogiri/xml/node_set.rb:186:in `upto'
/vendor/bundle/ruby/2.0.0/gems/nokogiri-1.6.6.2/lib/nokogiri/xml/node_set.rb:186:in `each'
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer/checks/links.rb:30:in `run'
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer.rb:121:in `block (2 levels) in check_files_for_internal_woes'
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer.rb:118:in `each'
/vendor/bundle/ruby/2.0.0/gems/html-proofer-2.1.0/lib/html/proofer.rb:118:in `block in check_files_for_internal_woes'
[...]

PR #179, opened 2015-02-20
Do a better job reporting failed attributes

gjtorikian:
Will close https://github.com/gjtorikian/html-proofer/issues/173.

If for some reason we cannot create a checker object--looks like if you have zany or broken attributes, this can be the case--we will report it as a failure. 

Perhaps there should be an option to toggle this off. For example, if you have `xmlns:whatever`, as @mathie experienced, we will just not create an attribute on the object, and we won't report the error, either. @penibelst what do you think?

### DIFF
--- a/lib/html/proofer/checks/favicon.rb
+++ b/lib/html/proofer/checks/favicon.rb
@@ -9,8 +9,13 @@ end
 class FaviconCheck < ::HTML::Proofer::CheckRunner
 
   def run
-    @html.xpath('//link[not(ancestor::pre or ancestor::code)]').each do |favicon|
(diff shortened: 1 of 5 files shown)


penibelst:
@gjtorikian Reporting is better than throwing an exception. The implementation looks good to me.

gjtorikian:
My concern is with this breaking existing sites ("But my attribute was valid a few days ago!"). I'm not super sure of merging this in just yet.

penibelst:
> My concern is with this breaking existing sites.

I understand. The #181 fixes #173 without breaking existing sites.

PR #180, opened 2015-02-22
Add support for validation of all svg elements.

mareksuscak:
Html proofer fails to validate svg elements like `<path>` `<defs>` and others so I thought it would be nice to add support for it too.


penibelst:
With this change Proofer will ignore all SVG elements. Am I right?

penibelst:
Maybe it’s better to remove SVG elements in https://github.com/mareksuscak/html-proofer/blob/master/lib/html/proofer/check_runner.rb#L55?

mareksuscak:
@penibelst Yes you are perfectly right - sorry I have missed that part that removes tags completely. I'm going to revert my commit and re-implement the check in check_runner.rb on line 55.

mareksuscak:
@penibelst Ok seems like svg can't be queried using a CSS selector when it has a namespace defined. I'll try to use XPath.

mareksuscak:
@penibelst I have managed to fix my failing test somehow but seems like I have broken something else. 

Could you please check my solution and help me to sort this out? I think I found out a bug in the code - nokogiri validates the document immediately after the file is loaded so when you want to unlink part of the document to prevent validation of that specific part you have to force re-validation by reloading the modified HTML.

mareksuscak:
Ok got it. nokogiri's to_s function returns formatted HTML so the documents get reformatted and thus cause a lot of tests to fail. Due to nokogiri's drawback that is mentioned in my previous comment I would suggest to revert back to my first solution because I don't see any other solution for that at moment. What's your opinion?

penibelst:
Maybe you try:

```ruby
html.xpath('//code | //pre | //svg').each(&:unlink)
```

mareksuscak:
But like I said in my recent comments. The problem finally wasn't that SVG can not be selected using css selectors - it was selected & removed correctly but the document still contained errors because it gets validated immediately after it's been loaded and removal of any element doesn't change anything.

penibelst:
Ah, I understand now.

mareksuscak:
Fine. So what's your suggestion then? Revert to my very first solution?

penibelst:
> Revert to my very first solution?

@mareksuscak Yes, please revert to your first draft. This solves your problem. Sorry for misleading.

I [asked](https://github.com/sparklemotion/nokogiri/issues/1255) Nokogiri people to help with this issue. Maybe we can figure out a better solution.

mareksuscak:
Ok so what now will you accept the pull request or ask on the mailing list?

penibelst:
@mareksuscak Your fix will appear in the next version. Thank you.

Issue #192, opened 2015-03-01
Problems with --href-swap 


eksperimental:
i found a real case that i need to use that has several issue when trying to set --href-swap from the command line
` --href-swap "/docs/:http://website.com/docs/" `
1st, it will take it as a regular expression because its surrounded by //, and there is not way to escape it properly, without giving up the original intention,
2nd, the colon, escaping doesn't work and it splits the strings into whatever amount of colons are there (I'll be submitting a patch for this).

i tried to hack around a little bit but I couldn't do much.
Maybe it escaping need to be implemented, 
also assigning regular expressions only with literals to avoid confusion with a situation like the one above.

gjtorikian:
Fixed by https://github.com/gjtorikian/html-proofer/pull/193.

PR #193, opened 2015-03-01
only split into two

eksperimental:
this avoids the problem when there's more than one colon in the string, to split it into more pieces

around the same code, there is a situation when the string is surrounded by // that needs to be solved
it is converted into a regular expression, and then the split method is not available and the application breaks.

thank you


eksperimental:
the original issue was #192

gjtorikian:
thanks!

PR #196, opened 2015-03-20
Add --ignore-empty-alt option

wolfgang42:
I'm trying to use `htmlproofer` on a documentation site with a lot of screenshots. The screenshots don't have any information that's missing from the text, so they all have `alt=""` set. Unfortunately this causes htmlproofer to complain "`image [...] does not have an alt attribute`".
#39 raised this problem, but resolved it with the `alt_ignore` option; I don't want to have to update the list every time I write more documentation.

This code is probably terrible; I don't know Ruby very well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ The `HTML::Proofer` constructor takes an optional hash of additional options:
 | Option | Description | Default |
 | :----- | :---------- | :------ |
 | `alt_ignore` | An array of Strings or RegExps containing `img`s whose missing `alt` tags are safe to ignore. | `[]` |
+| `alt_empty_igore` | If `true`, don't complain about images with empty alt tags. | `false` |
 | `check_external_hash` | Checks whether external hashes exist (even if the website exists). This slows the checker down. | `false` |
 |`checks_to_ignore`| An array of Strings indicating which checks you'd like to not perform. | `[]`
 | `directory_index_file` | Sets the file to look for when a link refers to a directory. | `index.html` |
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ The `HTML::Proofer` constructor takes an optional hash of additional options:
 | Option | Description | Default |
 | :----- | :---------- | :------ |
 | `alt_ignore` | An array of Strings or RegExps containing `img`s whose missing `alt` tags are safe to ignore. | `[]` |
-| `alt_empty_igore` | If `true`, don't complain about images with empty alt tags. | `false` |
+| `alt_empty_ignore` | If `true`, don't complain about images with empty alt tags. | `false` |
 | `check_external_hash` | Checks whether external hashes exist (even if the website exists). This slows the checker down. | `false` |
 |`checks_to_ignore`| An array of Strings indicating which checks you'd like to not perform. | `[]`
 | `directory_index_file` | Sets the file to look for when a link refers to a directory. | `index.html` |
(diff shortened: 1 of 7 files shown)


LaurentGoderre:
:+1:

gjtorikian:
The code's not bad, but it's missing tests. I'd really rather this be merged into the existing `alt_ignore` check, though.

I've logged an issue against myself to do that: https://github.com/gjtorikian/html-proofer/issues/203

gjtorikian:
> I've logged an issue against myself to do that

Unless, of course, you want to tackle it in this PR!

wolfgang42:
> The code's not bad, but it's missing tests.

@gjtorikian Which tests is it missing? I thought I covered all the cases in https://github.com/wolfgang42/html-proofer/commit/8836d897f786f63904ca363d40b5225f844cd48d#diff-087e8e4b3841eae80cc535cba9fc4a33R99

>  I'd really rather this be merged into the existing alt_ignore check, though.

I'm not sure how you'd do that--the alt_ignore check is looking at the src attribute, not the alt attribute. Unless I've misunderstood what you mean?

gjtorikian:
If `alt_ignore` were updated to ignore `alt=""`, then we wouldn't need a new option. You could simply do `{:alt_ignore => [/.*/]})`--that is, ignore the `alt` tags on *any* image, since `/.*/` matches every `src`. 

I think the fact that `alt_ignore` doesn't skip zero-length `alt` attributes is a bug that should be rectified.

wolfgang42:
> I think the fact that `alt_ignore` doesn't skip zero-length `alt` attributes is a bug that should be rectified. 

I was under the impression that it was a feature: there's [a test specifically checking that `it 'fails for image with an empty alt attribute'`](https://github.com/gjtorikian/html-proofer/blob/2f7f0bd10cff157b3f16c725a40a682ef4d5c947/spec/html/proofer/images_spec.rb#L16). Updating `alt_ignore` to ignore `alt=""` would break this test, should I still do it?

gjtorikian:
I feel like we're not communicating well. I also just realized that what this PR is trying to accomplish is already handled by html-proofer. I've opened https://github.com/gjtorikian/html-proofer/pull/205 to demonstrate what I'm talking about:

* Right now, without any options, html-proofer will complain if an image tag is missing the `alt` attribute, or if the `alt` attribute is empty.
* There is a flag called `alt_ignore` that matches the `src` attribute of `img` tags.
* If you set `alt_ignore` to ignore a certain `src` filename, then any issues with the `alt` tag is ignored.

So, instead of a new "ignore blank `alt` tags," it's more apt to say, "ignore the `alt` tags for images with the following `src`s." And, if you have a bunch of `img` tags with empty `alt`s, you can just define the `src` tag to ignore to be `/.*/`--which expands to "everything." It's an existing configuration option that can be used to achieve what you need. 

Anyway, https://github.com/gjtorikian/html-proofer/pull/205 improves upon the existing code, because it calls `strip` to remove all the whitespace from a tag, so that `alt="     "` also fails.

wolfgang42:
>So, instead of a new "ignore blank alt tags," it's more apt to say, "ignore the alt tags for images with the following srcs." And, if you have a bunch of img tags with empty alts, you can just define the src tag to ignore to be /.*/--which expands to "everything." It's an existing configuration option that can be used to achieve what you need. 

That doesn't do what I need, though. Passing `--alt-ignore '/.*/'` causes `alt=""` to be ignored, yes, but it also ignores *missing* `alt`s, which I still want to catch. (That is, I want to fail on `` but pass ``.)
As I explain in the OP, I have a lot of screenshots which don't add anything for non-visual readers. I want to make sure they *have* an `alt` attribute, so the screen reader doesn't read off the filename, but the attribute value needs to be *blank* to indicate that the image does not add any additional information.

gjtorikian:
Okay. I understand now. I'm sorry for misunderstanding before. I didn't quite comprehend how the screenreader was working.

I've introduced a merge conflict here after merging https://github.com/gjtorikian/html-proofer/pull/205. I've pulled the commits in this branch into another PR so that you retain authorship credit: https://github.com/gjtorikian/html-proofer/pull/207

In #207 I'll make the changes I want to see to the tests and code as it's easier/quicker for me to push to the new branch. Closing this one out!

Thanks for making the PR, and sorry it's taking me so long to get it.

Issue #200, opened 2015-04-19
failed: 0 Server returned nothing (no headers, no data)


cornernote:
Some external sites are giving an error when I test from travis-ci.org:

*  External link http://dom.as/2009/02/03/mydumper/ failed: 0 Server returned nothing (no headers, no data)
*  External link https://wordpress.com/ failed: 0 Server returned nothing (no headers, no data)

I added them to my exclude, but I'd rather not.  Any ideas as to why this happens?

benbalter:
I am also seeing this with TechCrunch URLs (in addition to all WordPress URLs). See https://travis-ci.org/benbalter/benbalter.github.com/builds/59152824 for an example. Perhaps Travis's IP is blacklisted? Curl or requesting the page without an agent header returns the content as expected.

gjtorikian:
> Perhaps Travis's IP is blacklisted?

Seems not. I see `curl` working too, but proofer fails locally:

``` ruby
HTML::Proofer.new(['http://techcrunch.com/2011/01/07/twitter-informs-users-of-doj-wikileaks-court-order-didnt-have-to/']).run

Checking 1 external link...
- 
  *  External link http://techcrunch.com/2011/01/07/twitter-informs-users-of-doj-wikileaks-court-order-didnt-have-to/ failed: 0 Server returned nothing (no headers, no data)
RuntimeError: HTML-Proofer found 1 failure!
```

gjtorikian:
The Typhoeus User Agent is being blocked:

``` ruby
HTML::Proofer.new(['http://techcrunch.com/2011/01/07/twitter-informs-users-of-doj-wikileaks-court-order-didnt-have-to/'], :typhoeus => { :verbose => true, :headers => { 'User-Agent' => 'LOL' }}).run

HTML::Proofer.new(['http://techcrunch.com/2011/01/07/twitter-informs-users-of-doj-wikileaks-court-order-didnt-have-to/'], :typhoeus => { :verbose => true, :headers => { 'User-Agent' => 'LOL' }}).run
Running ["ImageCheck", "LinkCheck", "ScriptCheck"] checks on ["http://techcrunch.com/2011/01/07/twitter-informs-users-of-doj-wikileaks-court-order-didnt-have-to/"] on *.html... 


Checking 1 external link...
Hostname was NOT found in DNS cache
  Trying 192.0.79.33...
Connected to techcrunch.com (192.0.79.33) port 80 (#0)
[...]

Issue #197, opened 2015-04-03
403 error for drupal.org domain


kostajh:
Any idea why `html-proofer` is returning "403 No error" for `drupal.org` links? See for example the output from https://travis-ci.org/savaslabs/savaslabs.github.io/builds/57088282

Other `https` links on that site work just fine.

gjtorikian:
Such a link is behind authentication; when I visit it, I get https://www.drupal.org/user/login?destination=user/3013749. 

I had some ideas around this: https://github.com/gjtorikian/html-proofer/issues/86 But for now if it's an annoyance I would suggest setting up the `href_ignore` option to skip over these links. :cry:

gjtorikian:
Also closing as a dupe of #86

kostajh:
@gjtorikian thanks, that makes sense. I didn't even realize that the `/u/{username}` links on drupal.org required authentication - the same page (aliased to a different URL) loads fine for unauthenticated user.

What's odd, though, is that the last two errors in [the build](https://travis-ci.org/savaslabs/savaslabs.github.io/builds/57088282) definitely do not require authentication but are yielding 403 errors:

- External link https://www.drupal.org/ failed: 403 No error
- External link https://www.drupal.org/project/drupalgeddon failed: 403 No error

If you have thoughts on those two please let me know.

gjtorikian:
Now that is indeed interesting. :smile:

benbalter:
I'm seeing this on several sites, including `codex.wordpress.org`. I wonder if travis's IP is blacklisted?  See https://travis-ci.org/benbalter/benbalter.github.com/builds/59152824 for an example.

benbalter:
Both this and #200 (which may be related) seem domain based. Perhaps a test-wide domain blacklist?

gjtorikian:
This is definitely related to #200. Compare

```
curl -A 'Typhoeus' -I https://www.drupal.org/
HTTP/1.1 403 Forbidden
Cache-Control: no-cache
```

to

```
 curl -I https://www.drupal.org/
HTTP/1.1 200 OK
Age: 1640
```

benbalter:
@gjtorikian you're 100% right. Reached out to friends at Drupal and WP to see what's up.

gjtorikian:
I appreciate that. Still, the problem seems to expand beyond just the scope of those large providers:

```
curl -I http://www.law.cornell.edu/supct/html/92-1292.ZS.html
[...]

PR #201, opened 2015-04-20
Set default user agent

benbalter:
This pull request sets the default user agent as `Mozilla/5.0 (compatible; HTML Proofer/2.1.0}; +https://github.com/gjtorikian/html-proofer)`. The user agent screen is based off of GoogleBot's user agent format. This should allow us to bypass Typhoeus-specific blocks on WordPress and Drupal.

Via @gjtorikian over in https://github.com/gjtorikian/html-proofer/issues/200#issuecomment-94313702:

> What are the ethics involved in my changing the User-Agent to something non-static to avoid being blacklisted again?

I believe as long as we're honest, and identify as HTML-Proofer (which can be blocked if that is in fact the desired behavior), we're fine. I don't think we should e.g., have a random string as a user agent hash. I suspect the services are blocking Typhoeus to prevent scraping, which is not what we're doing here.

Fixes https://github.com/gjtorikian/html-proofer/issues/197. Fixes https://github.com/gjtorikian/html-proofer/issues/200.

### DIFF
--- a/lib/html/proofer.rb
+++ b/lib/html/proofer.rb
@@ -44,7 +44,10 @@ module HTML
       }
 
       @typhoeus_opts = opts[:typhoeus] || {
-        :followlocation => true
+        :followlocation => true,
+        :headers => {
+          "User-Agent" => "Mozilla/5.0 (compatible; HTML Proofer/#{VERSION}; +https://github.com/gjtorikian/html-proofer)"
+        }
       }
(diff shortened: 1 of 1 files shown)


gjtorikian:
Thanks man. I'll talk about the test failure over in https://github.com/gjtorikian/html-proofer/pull/202.

PR #202, opened 2015-04-20
Faster Travis builds

benbalter:
* Use the container-based infra which usually has a shorter queue 
* Cache gems between builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,6 @@ env:
 addons:
   code_climate:
     repo_token:
+
+sudo: false
+cache: bundler


parkr:
Looks like timclem.wordpress.com has moved: http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/

gjtorikian:
We actually want it to be moved; it's been like that since The Beginning, since the test tests that 301 Redirects are followed.

This looks like another case of the user-agent being flagged:

```
curl -A 'Typhoeus' -I http://timclem.wordpress.com/2012/03/01/mind-the-end-of-your-line/ 
curl: (52) Empty reply from server
```

@benbalter If you merge master in here that might solve it.

benbalter:
Failing test is unrelated (continues to fail on master). I believe it's https://github.com/gjtorikian/html-proofer/issues/200, which wasn't completely fixed by the user agent.

PR #204, opened 2015-04-20
Better Typhoeus option handling

gjtorikian:
@benbalter Ugh so dumb. Since [the test](https://github.com/gjtorikian/html-proofer/blob/14edbcaaee5dc078f1d4c7ae938e553174741ba7/spec/html/proofer/links_spec.rb#L79-L83) was passing in its own Typhoeus options, it overwrote the User-Agent option. 

This PR changes things to keep the UA change while still respect other key changes.

Closes https://github.com/gjtorikian/html-proofer/issues/200.

### DIFF
--- a/lib/html/proofer.rb
+++ b/lib/html/proofer.rb
@@ -23,6 +23,13 @@ module HTML
 
     attr_reader :options, :typhoeus_opts, :hydra_opts, :parallel_opts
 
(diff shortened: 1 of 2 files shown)


benbalter:
We might need a deep merge type method in the long run (e.g., what if I pass a custom header), but nice catch.

PR #212, opened 2015-04-30
Make href_ignore work on all elements

### LINKED ISSUE #211: Exclude references via RegEx?
Hi,

is it possible to exclude some references via RegEx? It seems `file_ignore` only is for files and `href_ignore` for `href` attributes. 

But there’s no way to achieve the same for `src` attributes. Is that possible though or are you able to add this?

Thanks for your help in advance,
Anselm

penibelst:
Fix #211 

From what I feel `href_ignore` should work on all checkable elements. I don’t think we want dedicated ignores for every possible attribute, like `src_ignore`, `srcset_ignore`, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ The `HTML::Proofer` constructor takes an optional hash of additional options:
 | `href_ignore` | An array of Strings or RegExps containing `href`s that are safe to ignore. Note that non-HTTP(S) URIs are always ignored. | `[]` |
 | `href_swap` | A hash containing key-value pairs of `RegExp => String`. It transforms links that match `RegExp` into `String` via `gsub`. | `{}` |
 | `only_4xx` | Only reports errors for links that fall within the 4xx status code range. | `false` |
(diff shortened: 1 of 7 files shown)


gjtorikian:
Wouldn't `href_ignore` need to be renamed, then?

penibelst:
You are right. Any suggestions?

gjtorikian:
I do not, but I would like to think of one before this merges in. Perhaps `url_ignore`? The same could be said of `href_swap`. 

I also don't want to have to do a Major version bump just for this, so maybe this is being pedantic. The README for sure would need to be updated to note this change in behavior, though.

penibelst:
@gjtorikian Is this the right direction, so far?

gjtorikian:
Yeah, I think that makes more sense. Feel free to merge when you're happy with it. I'll open an issue to make note that we should drop `href_ignore` in a future major release.

penibelst:
I will finish it the upcoming week-end.

PR #222, opened 2015-06-07
Removed Codeclimate

penibelst:
Looks like we don’t use Codeclimate.


gjtorikian:
Heh. I just never bothered to learn how to set it up. Was it supposed to do something automatically?

penibelst:
Code Climate detects test coverage and code complexity on every pull request. This helps to improve the overall code quality.

PR #224, opened 2015-06-27
Wikipedia sends 301 now

penibelst:
This should fix the failures for the sorting test https://travis-ci.org/gjtorikian/html-proofer/jobs/68576491


gjtorikian:
It sends a 301 for missing items? That's.... weird.

PR #225, opened 2015-06-29
Create a test site on GitHub Pages

penibelst:
As we see in #224 external links are not reliable. I propose to create a test site for our external links and use them exclusively in our test suite.

The site should live in the `gh-pages` branch, at the address http://gjtorikian.github.io/html-proofer/ All files are as small as possible to accelerate our tests.

Should I upload these files to the `gh-pages` branch?


gjtorikian:
I'm not sure I understand what the advantage here is. We would having a live site gain us that the fixture files wouldn't? At the least I can see *more* work, as feature changes would need to go into `master`, and test changes would go into `gh-pages`.

penibelst:
The site is for external links only, a little set of defined resources: some 200s, some 301s and some 404s. They would replace over 30 external links we use now in the specs.

## Example

This fixture:
```html
<html>

<body>

<a href="http://www.asdo3IRJ395295jsingrkrg4.com">broken link!</a>
</body>

</html>
```

become

```html
<html>

<body>

<a href="http://gjtorikian.github.io/html-proofer/404.html">broken link!</a>
</body>

</html>
```

gjtorikian:
Oh, I see. You mean that we'll be able to control the status codes of the fixtures, because we'll be controlling the site.

Maybe we should use VCR instead to mock the status codes and replies. I think going this route would require complicated methods to keep the tests on `gh-pages` in sync with new features.

penibelst:
You are right, [VCR](http://www.relishapp.com/vcr/vcr/docs) looks more promising.

gjtorikian:
I don't mind setting up a PR later this week (just to avoid conflicts between the two of us doing it at the same time). 

Thanks for your dedicating in finding and fixing these kinds of issues. :heart:

PR #230, opened 2015-07-11
Add the url_ignore option to the binary

### LINKED ISSUE #229: url_ignore not available through the CLI
The `url_ignore` flag is not available through the CLI, but I believe it should be.

penibelst:
This should fix #229


gjtorikian:
Thanks!

Issue #233, opened 2015-07-31
Ignore script templates


akshayrawat:
Is it possible to ignore script templates just like HTML5 tags which aren't supported by Nokogiri/libxml2? Or is there another workaround to ignore errors on these snippets?

```html
<script type="text/html">
   <li></li>
</script>

<script type="text/x-handlebars-template">
   <li></li>
</script>
```

Error:

>  *  Element script embeds close tag

Its very common for frameworks ([Ember](http://guides.emberjs.com/v1.10.0/templates/handlebars-basics/#toc_defining-templates), [Knockout](http://knockoutjs.com/documentation/template-binding.html#note-1-rendering-a-named-template) etc.) to have html fragments as script text templates.

gjtorikian:
You can add the `data-proofer-ignore` attribute to any tag to ignore it: https://github.com/gjtorikian/html-proofer/blob/master/README.md#ignoring-content

Would that work?

akshayrawat:
I tried a few things, and that doesn't work

```html
<script type="text/html" id="navbar-logged-in" data-proofer-ignore>
    <li><a href="https://www.clerky.com/features">Home</a></li>
</script>
```

and

```html
<script type="text/html" id="navbar-logged-in" data-proofer-ignore>
    <li><a href="https://www.clerky.com/features" data-proofer-ignore>Home</a></li>
</script>
```

penibelst:
@akshayrawat What are the options you pass to Proofer?

akshayrawat:
```ruby
task :html_proofer do
  HTML::Proofer.new(
    "./build",
    {
      check_favicon: true,
      check_html: true #Yep

      parallel: {
        in_processes: 3
      }
    }
  ).run
end
```

gjtorikian:
Well hot diggity. I can reproduce this.

publicarray:
Has something changed? I received `Element script embeds close tag (line 129)` with `htmlproof ./_site --check-html --ignore-script-embeds --disable-external --url-ignore "#"` and with
```
<script data-proofer-ignore>
window.onload = function init() {
    var quiz1 = new dragDropQuiz({
        trailMode: false,
        alertResult: true,
        infoWrong: 'You can catch up on the Solar System <a href="/solar-system.html">here</a>.',
    });
};
</script>
```

PR #234, opened 2015-08-05
Ignore embedded scripts when asked

gjtorikian:
Closes https://github.com/gjtorikian/html-proofer/issues/233.

This PR proposes to add a new option, `ignore_script_embeds`, to resolve errors that read `Element script embeds close tag`.

I'm not 100% sure this is still the best way to handle this, but it might be the only way. From what I read in [Nokogiri's docs](http://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/SyntaxError), there's pretty much no information given from a parse error.

/cc @akshayrawat


lib/html/proofer/checks/html.rb
```diff
@@ -1,6 +1,9 @@
 # encoding: utf-8
 
 class HtmlCheck < ::HTML::Proofer::CheckRunner
+  class HtmlCheckable < ::HTML::Proofer::Checkable
+
+  end
```
penibelst: Why is this class `HtmlCheckable` here?

gjtorikian: Oh, I added it while testing something out. It's unnecessary.

akshayrawat:
It would be nice if html-proofer supported multiple html validators like [vnu](https://github.com/validator/validator). My current setup disables html validation via proofer and instead uses vnu for that.

gjtorikian:
> It would be nice if html-proofer supported multiple html validators like vnu. My current setup disables html validation via proofer and instead uses vnu for that.

That's a pretty good idea. It might be a pain to integrate a Java  executable into this Ruby library though.

spec/html/proofer/fixtures/html/ignore_script_embeds.html
```diff
@@ -0,0 +1,3 @@
+<script type="text/html" id="navbar-logged-in" data-proofer-ignore>
```
penibelst: I would remove `data-proofer-ignore` from this fixture. The spec “ignores embeded scripts when asked” must pass without the attribute.

geobrando:
Hey @gjtorikian.  Was wondering if there's anything currently in the way of merging this.

gjtorikian:
@geobrando Sorry! This is now out in 2.4.0: https://github.com/gjtorikian/html-proofer/releases/tag/v2.4.0

geobrando:
Thanks!

geobrando:
@gjtorikian 

I'm wondering if I'm missing something.  Either I don't know what I'm doing or there is something wrong with the tests.
```
  ruby:
    interpreter:  "ruby"
    version:      "2.1.2p95"
    date:         "2014-05-08"
    platform:     "i686-linux"
    patchlevel:   "2014-05-08 revision 45877"
    full_version: "ruby 2.1.2p95 (2014-05-08 revision 45877) [i686-linux]"
```
---

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  gem 'html-proofer', github: 'gjtorikian/html-proofer', :tag => 'v2.4.2'
end    
  
HTML::Proofer.new("./files",{:check_html => true, :ignore_script_embeds => true}).run
```
running with the fixture in `/files` yields

```
- ./files/repo-test.html
  *  Element script embeds close tag (line 2)
  *  Element script embeds close tag (line 2)
/home/geobrando/.rvm/gems/ruby-2.1.2/bundler/gems/html-proofer-f1f645f19a87/lib/html/proofer.rb:190:in `print_failed_tests': HTML-Proofer found 2 failures! (RuntimeError)
	from /home/geobrando/.rvm/gems/ruby-2.1.2/bundler/gems/html-proofer-f1f645f19a87/lib/html/proofer.rb:92:in `run'
	from /home/geobrando/Projects/htmlproofer-testing/test.rb:8:in `<main>'
```

geobrando:
Hey @gjtorikian. Is this working as intended?

colemanm:
I'm also still seeing the `Element script embeds close tag` errors when I have this new option enabled, running 2.5.1 latest. Same as @geobrando with `check_html` and `ignore_script_embeds` both true.

gjtorikian:
The documentation is wrong, but the feature is working. The options should be passed like this:

``` ruby
opts = { :check_html => true, :validation => { :ignore_script_embeds => true } }
```

That is, within the `:validation` namespace (because I anticipated future options).

geobrando:
@gjtorikian Ah.  You are correct.  Looking at the [binary](https://github.com/gjtorikian/html-proofer/blob/1aef3dcd9c1fec53b31bde99b7c63bc7c11473c9/bin/htmlproof), I think something needs to be changed to produce the proper namespaced option when using `--ignore-script-embeds` on  the CL but I couldn't figure out how to do it.  

From my limited knowledge of ruby it looks like you're mapping the CL flags directly to options that you pass to `HTML:proofer`.  I'll let you figure out how to do this as I assume you'd have to take into account adding additional options in the `:validation` namespace.

PR #239, opened 2015-09-01
correct ignore script option

geobrando:
Wasn't able to use `--ignore-script-embeds` on the command line.  This should fix that.


gjtorikian:
Ugh gross. Sorry! I really need to clean these options up. :frowning:

gjtorikian:
And thanks!

PR #264, opened 2015-11-09
Added --allow-hash-href option

### LINKED ISSUE #255: Empty hash in URL
Should be `<a href="#">` really be reported as a non-existing hash?

It is often used for resetting hash in sites where hash is used to some changes in content, e.g. tabs and AJAX-loaded content.

Workaround is this parameter: `--href-ignore "#"`

johnzeringue:
Fixes #255 by adding a more intuitive option to allow the `href` `#`.
This option, `--allow-hash-href`, is equivalent to `--href-ignore '#'`.
Its Ruby option is `allow_hash_href`, which should be a boolean and
defaults to `false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ The `HTML::Proofer` constructor takes an optional hash of additional options:
 
 | Option | Description | Default |
 | :----- | :---------- | :------ |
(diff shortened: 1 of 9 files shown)


gjtorikian:
Nice. That's pretty solid. Thanks!

penibelst:
Great.

PR #265, opened 2015-11-09
Add support for <base> tags

### LINKED ISSUE #166: HTML base tag
Hi,

I have not yet come up with a test for it, but I have the feeling html-proofer does not implement the [HTML base tag](http://www.w3schools.com/tags/tag_base.asp).

I myself only discovered the existence of the tag recently, which is really useful for not hardcoding too much in your site's URL.

Do you want to comment on that ?

Cheers, B.

bkeepers:
From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base)

> The **HTML `<base>` element** specifies the base URL to use for all relative URLs contained within a document.

This adds support for the base tag for links and images. For some reason the image tests are failing and I'm having some trouble figuring out why.

Fixes #166

### DIFF
--- a/lib/html/proofer/checkable.rb
+++ b/lib/html/proofer/checkable.rb
@@ -9,7 +9,7 @@ module HTML
 
       attr_reader :line
 
-      def initialize(obj, check, base = nil)
+      def initialize(obj, check)
         obj.attributes.each_pair do |attribute, value|
           instance_variable_set("@#{attribute.tr('-:.', '_')}".to_sym, value.value)
         end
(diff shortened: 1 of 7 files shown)


lib/html/proofer/checkable.rb
```diff
@@ -30,7 +30,11 @@ def initialize(obj, check)
       end
 
       def url
-        @src || @srcset || @href || ''
+        url = @src || @srcset || @href || ''
+        if @check.base
```
gjtorikian: For the tests, this can change to

``` ruby
if @check && @check.base
```

The reason `@check` is `nil` is because of the tests in *spec/checkable_spec.rb*. Technically you can parse HTML without associating it with a check to run.

spec/html/proofer/fixtures/images/relativeWithBase.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <base href="/links/">
```
gjtorikian: It seems like the `base` calculation is relative to the location of the file being tested:

```
"html-proofer/spec/html/proofer/fixtures/images/links/gpl.png"
"html-proofer/spec/html/proofer/fixtures/images/links/folder/assets/barrel.png"
"html-proofer/spec/html/proofer/fixtures/images/images/gpl.png"
"html-proofer/spec/html/proofer/fixtures/images/images/gpl.png"
```

I don't know if that's correct or not, but it's definitely why the test is failing here.

gjtorikian:
Closing as stale, but will explore fixing this soon.

PR #269, opened 2015-11-19
Regex example for clarity

plaindocs:
I spent a few mins looking up how to exclude URLs with regex, although probably people more familiar with Ruby won't have that issue.


gjtorikian:
Looks good to me!

gjtorikian:
Thanks! :tangerine:

Issue #267, opened 2015-11-15
htmlParseEntityRef throws error for valid iframe src URLs


jackfrancis:
Using the `check_html` option, I can't seem to get htmlproof to accept an iframe src URL with more than 1 param (i.e., a src string that will have a literal `&` to delimit param names). E.g.:

```
  *  htmlParseEntityRef: expecting ';' (line 92)
```

The relevant line 92:

```
<iframe frameborder="0" style="border:0" src="https://www.google.com/maps/embed/v1/place?q=1035%20Market%20Street%2C%20Suite%20550%20San%20Francisco%2C%20CA%2094103&key=R-E-D-A-C-T-E-D"></iframe>
```

The above code is obviously valid (excepting the redacted key value), and comes straight from Google:

https://developers.google.com/maps/documentation/embed/start

In the interim, I am using the `file_ignore` option to get HTML validation on the rest of my site, but I would love to be able to validate the above code. Thanks!

Joshua-Anderson:
Seeing the same problem with this `onClick` attribute of a link:

```
onClick="window.open('https://calendar.google.com/calendar/embed?src=o3uckm13gkcc5tu4u3846ik0pc%40group.calendar.google.com&ctz=America/Los_Angeles','Team Calendar','')"
```

Changing the google calendar url to 'https://test.com' causes the proofer to pass

bascht:
@gjtorikian I might have found a regression of this one + #386:

```html
<!-- will yield an error: *  822:87: ERROR: htmlParseEntityRef: expecting ';' (line x) -->
<script src="//maps.googleapis.com/maps/api/js?callback=boing&key=xyz"></script>

<!-- will pass: -->
<script src="https://maps.googleapis.com/maps/api/js?callback=boing&key=xyz"></script>
```

Should I open a new issue?

gjtorikian:
@bascht Yes please!

PR #273, opened 2015-11-24
Clean content before parsing with Nokogiri

gjtorikian:
Apparently, unescaped forward slashes or ampersands will cause Nokogiri to break. In these cases, it's best to just rewrite the content so that Nokogiri is happy.

I thought about performing this rewrite for only HTML validation cases, but it felt better to just always do a clean.

Closes https://github.com/gjtorikian/html-proofer/issues/267.


Joshua-Anderson:
Thank you! :heart:

benbalter:
I believe this may now prevent `government.github.com` from validating. See https://github.com/github/government.github.com/pull/423.

Issue #219, opened 2015-05-26
Introduce url_swap as companion for url_ignore


penibelst:
The new option `url_ignore` needs a companion that transforms links in every attribute. It should be an extended version of `href_swap`. I propose the name `url_swap`.

Related to #178, #214.

mdotasia:
Do you have road map for `url_ignore`?

penibelst:
@mdotasia `url_ignore` is available since version 2.3.0.

nasht00:
Reading this, do I understand that currently, `href_swap` does not replace in every element?

What's the current workaround to make it work with images, and script sources?

penibelst:
@nasht00, yes `href_swap` does replace only in elements with a `href` attribute. Therefore it’s for `a` elements only. The current workaround is to ignore those URLs you would swap.

nasht00:
Every single URL on my site needs to be swapped (they all start with a virtual `/subdomain/` which does not really exist in the file system).

nasht00:
Is there a roadmap for `url_swap`?

penibelst:
@nasht00 No, there is no roadmap.

gjtorikian:
Closed by HTMLProofer 3.0: https://github.com/gjtorikian/html-proofer/releases/tag/v3.0.0

Issue #214, opened 2015-05-06
Drop `href_ignore` in a future release


gjtorikian:
See https://github.com/gjtorikian/html-proofer/pull/212.

gjtorikian:
Closed by HTMLProofer 3.0: https://github.com/gjtorikian/html-proofer/releases/tag/v3.0.0

PR #287, opened 2015-12-23
Drop href ignore/swap in favor of url ignore/swap

gjtorikian:
Closes https://github.com/gjtorikian/html-proofer/issues/219, closes https://github.com/gjtorikian/html-proofer/issues/214.

The URL versions are more generic, and should act on more than just `href` elements.


gjtorikian:
Bah, wrong head.

Issue #278, opened 2015-12-05
Rename to `html-proofer`


gjtorikian:
I think the `html/proofer` style was a bad idea.

gjtorikian:
Closed by HTMLProofer 3.0: https://github.com/gjtorikian/html-proofer/releases/tag/v3.0.0

Issue #246, opened 2015-09-18
Rename `htmlproof` to `htmlproofer`


gjtorikian:
This will be a breaking change.

penibelst:
Do you want to talk about the future API changes?

gjtorikian:
Sure, open issues if you have strong feelings.

gjtorikian:
Closed by HTMLProofer 3.0: https://github.com/gjtorikian/html-proofer/releases/tag/v3.0.0

PR #290, opened 2015-12-23
Restructure classes and binary program

gjtorikian:
Closes https://github.com/gjtorikian/html-proofer/issues/278, closes https://github.com/gjtorikian/html-proofer/issues/246.

The whole `html/proofer` style was taken straight off of https://github.com/jch/html-pipeline. It's pretty un-Ruby and in fact very Perl. I think I'd prefer to go back to something more Ruby-like.

This PR renames everything to move to a top-level `HTMLProofer` class, rather than nested modules. It also renames the binary to better reflect the program (`htmlproofer`, rather than `htmlproof`).


gjtorikian:
Guh, wrong head AGAIN.

PR #221, opened 2015-06-02
Checkable#file_path: allow for dropping html extension.

parkr:
Using a new feature in Jekyll 3 which allows you to drop the `.html` in the links. Yields this error:

```text
- ./_site/go/index.html
  *  internally linking to /go/jekyll-build-server, which does not exist (line 50)
  *  internally linking to /go/merge-pr, which does not exist (line 58)
HTML-Proofer found 2 failures!
```

This is a fix for this new feature coming in Jekyll 3.

Maybe related to #195? /cc @gjtorikian @penibelst

PR #311, opened 2016-03-02
Allow extensionless URLs

### LINKED ISSUE #309: HTML pretty links & extensions
I'm using html-proofer with Jekyll, and both Jekyll's built-in web server and github pages will serve `/example.html` as `/example`, to make the links prettier. However, because html-proofer operates on the files, not using a server with that ability, I have to include `.html` in my links to make it pass. Is there a way, or could there *be* a way, to get html-proofer to check an .html extension if there wasn't one specified? This would of course have to be non-default behaviour, otherwise it could be glossing over broken links if the server *wasn't* doing that job.

Floppy:
To support Jekyll 3 and github pages, which automatically deliver `/example.html` if `/example` is requested. Code is cribbed from @parkr's code in #221. I'm not finished with this yet, though it works and I'm using it to test my Jekyll 3 site. 

Still to do is to make this optional behaviour, so it doesn't affect existing users unless they turn it on.

@parkr @gjtorikian, any insight you might have into why #221 was dropped (and therefore what might need fixing to get it accepted) would be lovely.

Resolves #309.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ The `HTML::Proofer` constructor takes an optional hash of additional options:
 | :----- | :---------- | :------ |
 | `allow_hash_href` | If `true`, ignores the `href` `#`. | `false` |
 | `alt_ignore` | An array of Strings or RegExps containing `img`s whose missing `alt` tags are safe to ignore. | `[]` |
+| `assume_extension` | Automatically add extension (e.g. `.html`) to file paths, to allow extensionless URLs (as supported by Jekyll 3 and Github Pages) | `nil` |
 | `check_external_hash` | Checks whether external hashes exist (even if the website exists). This slows the checker down. | `false` |
 | `check_favicon` | Enables the favicon checker. | `false` |
 | `check_html` | Enables HTML validation errors from Nokogiri | `false` |
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ The `HTML::Proofer` constructor takes an optional hash of additional options:
 | :----- | :---------- | :------ |
 | `allow_hash_href` | If `true`, ignores the `href` `#`. | `false` |
 | `alt_ignore` | An array of Strings or RegExps containing `img`s whose missing `alt` tags are safe to ignore. | `[]` |
-| `assume_extension` | Automatically add extension (e.g. `.html`) to file paths, to allow extensionless URLs (as supported by Jekyll 3 and Github Pages) | `nil` |
+| `assume_extension` | Automatically add extension (e.g. `.html`) to file paths, to allow extensionless URLs (as supported by Jekyll 3 and GitHub Pages) | `nil` |
 | `check_external_hash` | Checks whether external hashes exist (even if the website exists). This slows the checker down. | `false` |
 | `check_favicon` | Enables the favicon checker. | `false` |
 | `check_html` | Enables HTML validation errors from Nokogiri | `false` |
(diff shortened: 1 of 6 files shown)


Floppy:
Also, the test totally shouldn't pass right now as the fixture isn't in there. How odd.

Floppy:
OK, this doesn't work at all in fact, as is. But, I have a fix incoming.

Floppy:
OK, I'm happy with this. `assume_extension` option added, which specifies the extension to add (default `nil`). Tests that the option is required, and that it works with hash URLs and subfolders. Also added docs. Paging @gjtorikian for review :)

Floppy:
You can see this in use at https://github.com/SomethingNewUK/somethingnewuk.github.io/commit/248113a4f51329b981fa6b408a46c2afeeef5a62

README.md
```diff
@@ -156,6 +156,7 @@ The `HTML::Proofer` constructor takes an optional hash of additional options:
 | :----- | :---------- | :------ |
 | `allow_hash_href` | If `true`, ignores the `href` `#`. | `false` |
 | `alt_ignore` | An array of Strings or RegExps containing `img`s whose missing `alt` tags are safe to ignore. | `[]` |
+| `assume_extension` | Automatically add extension (e.g. `.html`) to file paths, to allow extensionless URLs (as supported by Jekyll 3 and Github Pages) | `nil` |
```
gjtorikian: Capital H GitHub. :wink:

lib/html/proofer/checkable.rb
```diff
@@ -107,11 +107,13 @@ def internal?
       def file_path
         return if path.nil?
 
+        path_dot_ext = path + @check.options[:assume_extension].to_s
+
         if path =~ %r{^/} # path relative to root
           base = File.directory?(@check.src) ? @check.src : File.dirname(@check.src)
-        elsif File.exist?(File.expand_path path, @check.src) # relative links, path is a file
+        elsif File.exist?(File.expand_path path, @check.src) || File.exist?(File.expand_path path_dot_ext, @check.src) # relative links, path is a file
```
gjtorikian: Mind adding the parenthesis here for both of the calls? Not sure why they were missing in the first place...

gjtorikian:
I like this a lot--thanks! Made a few minor comments.

Floppy:
both addressed :)

gjtorikian:
I'll be releasing this as part of 3.0 in a bit. Just a head's up that I change the code a bit.

`assume_extension` is now a boolean `true / false` (default to `false`). Given that the default extension is already `.html`, it didn't seem to make sense to provide a second extension. This way, the code can simply be:

``` ruby
HTML::Proofer.new("./_site", 
 typhoeus: {ssl_verifypeer: false, timeout: 30}, 
 url_ignore: ignored, 
 check_html: true, 
 assume_extension: true).run
```

Or, if you want a different extension:

``` ruby
HTML::Proofer.new("./_site", 
 typhoeus: {ssl_verifypeer: false, timeout: 30}, 
 url_ignore: ignored, 
 check_html: true, 
 extension: ".htm",
 assume_extension: true).run
```

Without this change, in that second case, the could would've looked like:

``` ruby
HTML::Proofer.new("./_site", 
 typhoeus: {ssl_verifypeer: false, timeout: 30}, 
 url_ignore: ignored, 
 check_html: true, 
 extension: ".htm",
 assume_extension: ".htm").run
```

which seemed excessive.

Floppy:
Yay! I did have that as the behaviour in an earlier version, so I'm cool with that :)


Thanks for merging!

> 
> 
> 
>

PR #319, opened 2016-03-14
htmlproofer should set status codes to ints

parkr:
`ignored_status_codes.include? 429` will be false if `ignored_status_codes = ["429"]`

### DIFF
--- a/bin/htmlproofer
+++ b/bin/htmlproofer
@@ -51,6 +51,8 @@ Mercenary.program(:htmlproofer) do |p|
       end
       options[option.config_key.to_sym] = opts[option.config_key]
     end
+    
+    options[:http_status_ignore] = options[:http_status_ignore].map(&:to_i)
 
     # some minor manipulation of a special option
     unless opts['url_swap'].nil?
--- a/bin/htmlproofer
+++ b/bin/htmlproofer
@@ -51,8 +51,6 @@ Mercenary.program(:htmlproofer) do |p|
       end
       options[option.config_key.to_sym] = opts[option.config_key]
     end
-    
-    options[:http_status_ignore] = options[:http_status_ignore].map(&:to_i)
 
     # some minor manipulation of a special option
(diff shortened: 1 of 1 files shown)


gjtorikian:
I like this. Thanks!

bin/htmlproofer
```diff
@@ -71,6 +71,8 @@ Mercenary.program(:htmlproofer) do |p|
 
     options[:cache] = {}
     options[:cache][:timeframe] = opts['timeframe'] unless opts['timeframe'].nil?
+    
+    options[:http_status_ignore] = Array(options[:http_status_ignore]).map(&:to_i)
```
gjtorikian: This will need to be `options[:http_status_ignore].split(',')` to get the array for multiple codes.

parkr: You specify it as an `Array` so OptionsParser will give it to you as such. I tested it locally and it works. :smile:

gjtorikian: Can you add a test for it? In `irb` at least I get:

```
Array("421, 422").map(&:to_i)
=> [421]
irb(main):005:0> 
```

https://github.com/gjtorikian/html-proofer/blob/4434b7eecb19d869c59c92e6480661d18866253c/spec/html-proofer/command_spec.rb is where all the `bin` tests go. :heart:

parkr: @gjtorikian the input isn't `"421, 422"`, it's `["421", "422"]` I'll add a test.

parkr: Scratch that. @gjtorikian I can't use both VCR and the executable because the executable doesn't share the same Ruby env.

Trust that [this line](https://github.com/gjtorikian/html-proofer/blob/4434b7eecb19d869c59c92e6480661d18866253c/bin/htmlproofer#L32) does what I'm saying it does?

gjtorikian:
Meh, I trust you. Sorry for the delay.

parkr:
❤️

PR #320, opened 2016-03-15
Added escape sequences for --url-swap

### LINKED ISSUE #293: --url-swap keys can't contain colon
Looking at [bin/htmlproof](https://github.com/gjtorikian/html-proofer/blob/master/bin/htmlproof), it's obvious that something like shown [here](http://tongueroo.com/articles/getting-html-proofer-to-work-with-canonical-url-for-google-seo/) is not possible from the command line, because the command line client simply splits each key-value pair on `:`, taking just the protocol from any URI key and prefixing the intended value.

I propose using `\` as an escape character and supporting `\:` and `\\` as escape sequences. This allows for colons and backslashes in both keys and values. The command line equivalent of the Ruby code in the article would then be `htmlproof --url-swap http\://tongueroo.com: ./_site`.

@gjtorikian, if you're okay with this solution, I'd be glad to make a pull request.

johnzeringue:
Fixed #293 by adding `\\` and `\:` as escape sequences for the
`--url-swap` flag. This allows command line users the same
expressiveness as Ruby users.

### DIFF
--- a/bin/htmlproofer
+++ b/bin/htmlproofer
@@ -37,7 +37,7 @@ Mercenary.program(:htmlproofer) do |p|
   p.option 'only_4xx', '--only-4xx', 'Only reports errors for links that fall within the 4xx status code range'
   p.option 'timeframe', '--timeframe <time>', String, 'A string representing the caching timeframe.'
   p.option 'url_ignore', '--url-ignore link1,[link2,...]', Array, 'A comma-separated list of Strings or RegExps containing URLs that are safe to ignore. It affects all HTML attributes. Note that non-HTTP(S) URIs are always ignored'
-  p.option 'url_swap', '--url-swap re:string,[re:string,...]', Array, 'A comma-separated list containing key-value pairs of `RegExp => String`. It transforms URLs that match `RegExp` into `String` via `gsub`.'
+  p.option 'url_swap', '--url-swap re:string,[re:string,...]', Array, 'A comma-separated list containing key-value pairs of `RegExp => String`. It transforms URLs that match `RegExp` into `String` via `gsub`. The escape sequences `\\:` and `\\\\` should be used to produce literal `:`s and `\\`s.'
 
   p.action do |args, opts|
     args = ['.'] if args.empty?
@@ -57,8 +57,37 @@ Mercenary.program(:htmlproofer) do |p|
     unless opts['url_swap'].nil?
       options[:url_swap] = {}
       opts['url_swap'].each do |s|
-        pair = s.split(':', 2)
-        options[:url_swap][Regexp.new(pair[0])] = pair[1]
+        re = string = ''
+
+        is_re = true
+        is_escaped = false
+
+        s.each_char do |c|
+          value = c
+
+          case c
+          when '\\'
+            value = '' unless is_escaped
+            is_escaped = !is_escaped
+          when ':'
+            if is_escaped
+              is_escaped = false
+            else
+              value = ''
+              is_re = false
+            end
+          else
+            is_escaped = false
+          end
+
+          if is_re
+            re += value
+          else
+            string += value
+          end
+        end
+
+        options[:url_swap][Regexp.new(re)] = string
       end
     end
 
--- a/README.md
+++ b/README.md
@@ -144,8 +144,9 @@ Pass in options through the command-line, like this:
 htmlproofer ./out --url-swap wow:cow,mow:doh --extension .html.erb --url-ignore www.github.com
 ```
 
-Note: since `url_swap` is a bit special, you'll pass in a pair of `RegEx:String` values.
-`htmlproofer` will figure out what you mean.
+Note: since `url_swap` is a bit special, you'll pass in a pair of `RegEx:String`
+values. The escape sequences `\:` and `\\` should be used to produce literal
+`:`s and `\`s. `htmlproofer` will figure out what you mean.
 
 ### Using with Jekyll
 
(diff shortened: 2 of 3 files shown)


johnzeringue:
@gjtorikian and @udiboy1209, here's a proof of concept. I'd love any feedback.

Having modified the test case to use the new escape sequences, I'm not crazy about the look of `\\\\\\\\` just to produce a single `\`. Of course, real command line users would "only" need `\\\\`, but I can't help but wonder if `$:` and `$$` wouldn't be a better syntax, since `\` is so prevalent in regular expressions.

From an engineering standpoint, I don't like how much parsing code sits inside `bin/htmlproofer` or how there's only one test for fairly tricky code. @gjtorikian, I was curious if/how you'd like any new files or tests.

udiboy1209:
@johnzeringue nice! Not so sure about using `$:` and `$$` on the command line because `bash` will look at them variables and try to substitute them, hence we will just end up using more escapes. 

On a different note, what do you think about passing two options to `--url-swap` instead of a combined one like so: `--url-swap "RegExp" "String"`. No extra escaping required for this I guess. This will break the current argument format but will save a lot of headache with escaping correctly.

johnzeringue:
You're absolutely right about `$`s. Since I commented, I also realized that it's also a common regex symbol. I still think finding another escape character is promising, we'll just need to cross-reference a couple different lists.

I'm not sure if your suggestion is possible with the CLI framework being used, although I'm not sure.

Any solution will break the current format in some way or lead us to use a different option name.

johnzeringue:
Space might not be such a bad delimiter, though. Similar to your suggestion, but instead having `--url-swap "pattern replacement"`. This is based on me assuming (with good reason) that there's never a use case for having a space in either pattern or replacement. The downside to this is that people who try `--url-swap pattern replacement` are going to be sad.

johnzeringue:
`%%` has the nice property of not being a regex symbol, a bash symbol, or a substring of a valid URL. For example, `--url-swap pattern%%replacement`. It's weird, but it makes the pattern and the replacement the most readable. I don't think any of `` `~!@#$%^&*()-_+=[{]}\|;:'",<.>/? `` have these properties on their own, which knocks out the standard US keyboard.

Maybe `::` is a better but similar alternative?

There are still some funny edge cases with this, though, like `--url-swap \+%%%20%20` and `--url-swap %20%%%%`. `::` should have similar problems.

udiboy1209:
I don't think `<>` are used that frequently in urls without being escaped. We could use something like `=>` as our delimiter. Has the added benefit of looking like ruby syntax.

PS I like the idea of using space as the delimeter, isn't entirely obvious but will do the job.

gjtorikian:
Admittedly, I never use the binary, and traditionally, it's been the least tests / most problematic part of HTMLProofer, heh. I welcome any new tests for it but don't expect them to be a part of this PR. 

I wouldn't mind space as a delimiter, either, but then how would multiple swaps be implemented? In pairs?

```
--url-swap a b c d
```

vs

```
--url-swap a::b c::d
```

The second seems nicer to me. I don't quite follow why `,` can't continue to be used, though.

johnzeringue:
Sorry this has gotten so unwieldy. Let me clarify my thoughts.

First, I don't think that

    --url-swap a b c d

is possible using [Mercenary](https://github.com/jekyll/mercenary).

I think the closest we can come to that is

    --url-swap "a b c d"

The current PR implements a *complete solution*, meaning that by using escape sequences, we are able to unambiguously specify any regex-replacement pair.

The downside to the current solution is that `\` is already used as an escape character in regular expressions, Bash, and Ruby, meaning that we can need as many as 4 `\`s in practice and 8 `\`s in testing to represent a single `\`. This might not be an issue if we're okay with `htmlproofer --url-swap "\\\\A/articles/([\\\\w-]+):\\\\1.html"` in the terminal.

An alternative to this might be to use another escape character like `$` or `%`. For example, `htmlproofer --url-swap "http$://example.com/articles/([\\w-]+)$$:\\1.html"`.

The other approach is to try and find a *perfect delimiter*, a delimiter that, unlike `:`, is not used in URLs, regular expressions, or Bash. As I explained above, I don't think it exists on a typical American keyboard. A space isn't bad, but it'll always require quotation marks around each pair.

`::` and other "double delimiters" might be good enough for practical use, but as shown above, they too introduce ambiguity and we'll end up restricting the user similar to the current syntax.

My vote is for either the current implementation with `\` as an escape or a similar solution with a different escape character. I just wanted to share my thoughts on alternatives and have a short discussion before we merged this. Instead, I think I overcomplicated things.

Hopefully, this clears things up.

johnzeringue:
@gjtorikian, I've rebased this PR against `master` again. Sorry I overcomplicated things here. To be clear, the PR still uses `:` to delimit pairs, but adds the `\:` and `\\` escape sequences for literal colons and backslashes.

bin/htmlproofer
```diff
@@ -57,8 +57,38 @@ Mercenary.program(:htmlproofer) do |p|
     unless opts['url_swap'].nil?
       options[:url_swap] = {}
       opts['url_swap'].each do |s|
-        pair = s.split(':', 2)
-        options[:url_swap][Regexp.new(pair[0])] = pair[1]
+        re = ''
+        string = ''
```
gjtorikian: Mind making this inline?

```` ruby
re = string = ''
```

gjtorikian:
Cool! Do you mind updating the `README` to explain the changes in delimiting pairs, as well as new ways to escape those special characters?

johnzeringue:
I've updated the `README` and the `--help` documentation. I also made the change you noted inline.

gjtorikian:
Thank you!

Issue #321, opened 2016-03-19
External link failed: response code 0 means something's wrong


zachwhaley:
When I run html-proofer locally, I get failures from the external link checks.  The failures are always the same `External link failed: response code 0 means something's wrong` but not on the same links; sometimes 13 links sometimes 23 links fail.

What is interesting is that I don't see this behavior on Travis CI.

```
Running ["LinkCheck", "ScriptCheck", "ImageCheck"] on ["./_site"] on *.html... 


Checking 32 external links...
Ran on 166 files!


- ./_site/2014/11/22/pound-for-pound/index.html
  *  External link http://www.tomlinsons.com/ failed: response code 0 means something's wrong
- ./_site/2015/01/26/february-news-2015/index.html
  *  External link http://www.zoomroomonline.com/austin-dog-training.html failed: response code 0 means something's wrong
- ./_site/2015/03/29/april-news/index.html
  *  External link http://austinheartvet.com/ failed: response code 0 means something's wrong
- ./_site/2015/05/04/may/index.html
  *  External link http://texasfarmersmarket.org/mueller/ failed: response code 0 means something's wrong
  *  External link http://us6.campaign-archive2.com/?u=ce82f17bc088ab9ad887edb8d&id=b31d7d780c&e=43054ebcf2 failed: response code 0 means something's wrong
- ./_site/2015/06/03/june/index.html
  *  External link http://eepurl.com/boYVTX failed: response code 0 means something's wrong
  *  External link http://www.handmade4hounds.etsy.com/ failed: response code 0 means something's wrong
- ./_site/2015/10/12/october-news/index.html
  *  External link https://www.facebook.com/events/1486089575035397/ failed: response code 0 means something's wrong
- ./_site/2015/11/05/november/index.html
  *  External link https://www.etsy.com/shop/dogsbythebay failed: response code 0 means something's wrong
- ./_site/about/index.html
  *  External link http://gpacentraltexas.us6.list-manage.com/subscribe?u=ce82f17bc088ab9ad887edb8d&id=5f26041b9d failed: response code 0 means something's wrong
  *  External link http://instagram.com/greyhoundscentraltexas?ref=badge failed: response code 0 means something's wrong
  *  External link http://www.randalls.com/IFL/Grocery/Good-Neighbor failed: response code 0 means something's wrong
  *  External link http://www.wooftrax.com/ failed: response code 0 means something's wrong
[...]

Issue #299, opened 2016-02-07
Recent Travis failures


lurch:
Hello,

We've got html-proofer integrated with https://github.com/raspberrypi/documentation but recently we're seeing an awful lot of Travis build failures, when it shouldn't be failing.
https://travis-ci.org/raspberrypi/documentation/builds
https://travis-ci.org/raspberrypi/documentation/pull_requests
(it seems to be failing when checking external links)

Not being familiar with html-proofer or the inner-workings of Travis, I dunno if this is a fault with html-proofer, or Travis? Anything we can do to mitigate the situation?

wjdp:
Having a quick look I found a lot of attempted links to `Readme.md` - probably rather than `.html` and problems checking links on http://elinux.org

```
- ./_build/configuration/config-txt.md
  *  External link http://elinux.org/RPI_safe_mode failed: response code 0 means something's wrong
  *  External link http://elinux.org/RPiconfig failed: response code 0 means something's wrong
  *  External link http://en.wikipedia.org/wiki/Coordinated_Video_Timings failed: response code 0 means something's wrong
  *  External link http://en.wikipedia.org/wiki/Extended_display_identification_data failed: response code 0 means something's wrong
  *  External link http://en.wikipedia.org/wiki/Phase-locked_loop#Clock_generation failed: response code 0 means something's wrong
- ./_build/configuration/device-tree.md
  *  External link http://elinux.org/images/4/48/Experiences_With_Device_Tree_Support_Development_For_ARM-Based_SOC's.pdf failed: response code 0 means something's wrong
- ./_build/configuration/raspi-config.md
  *  External link http://elinux.org/RPi_raspi-config failed: response code 0 means something's wrong
- ./_build/hardware/raspberrypi/gpio/README.md
  *  External link http://elinux.org/RPi_Low-level_peripherals failed: response code 0 means something's wrong
- ./_build/hardware/raspberrypi/spi/README.md
  *  External link http://elinux.org/BCM2835_datasheet_errata failed: response code 0 means something's wrong
  *  External link http://en.wikipedia.org/wiki/Serial_Peripheral_Interface_Bus failed: response code 0 means something's wrong
- ./_build/installation/installing-images/windows.md
  *  External link http://elinux.org/RPi_Easy_SD_Card_Setup failed: response code 0 means something's wrong
- ./_build/installation/noobs.md
  *  External link http://osmc.tv/ failed: response code 0 means something's wrong
[...]

PR #322, opened 2016-03-23
Show return message for status code 0

mlinksva:
Message can be populated and useful to know. I've only seen in execution environment I can't control (CI) and [want](https://github.com/github/choosealicense.com/pull/374) to see additional feedback. https://github.com/gjtorikian/html-proofer/issues/299#issuecomment-194754495 shows someone else's example of this. But apparently (#321) others see status code 0 when run locally and wonder what's wrong.

Unresolvable host is one status code 0 message that I can produce easily. I haven't added a test for this as an expected message because I haven't figured out how to have test output not recorded, or recorded with message -- 2nd time new test (not in this PR) is run without deleting recording, there is no message, so test fails.

### DIFF
--- a/lib/html-proofer/url_validator.rb
+++ b/lib/html-proofer/url_validator.rb
@@ -150,7 +150,7 @@ module HTMLProofer
       elsif response.timed_out?
         handle_timeout(href, filenames, response_code)
       elsif response_code == 0
-        handle_failure(href, filenames, response_code)
+        handle_failure(href, filenames, response_code, response.return_message)
       elsif method == :head
         queue_request(:get, href, filenames)
       else
@@ -192,8 +192,8 @@ module HTMLProofer
       add_external_issue(filenames, msg, response_code)
     end
 
-    def handle_failure(href, filenames, response_code)
-      msg = "External link #{href} failed: response code #{response_code} means something's wrong"
+    def handle_failure(href, filenames, response_code, return_message)
+      msg = "External link #{href} failed: response code #{response_code} means something's wrong; return message says '#{return_message}'"
       @cache.add(href, filenames, 0, msg)
       return if @options[:only_4xx]
       add_external_issue(filenames, msg, response_code)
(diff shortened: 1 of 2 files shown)


gjtorikian:
Thanks!

fabiendv:
I use the latest version (**3.10.2**), and nothing changed.
I have always these issues (whereas links are good):

```
  *  External link https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css failed: response code 0 means something's wrong.
             It's possible libcurl couldn't connect to the server or perhaps the request timed out.
             Sometimes, making too many requests at once also breaks things.
             Either way, the return message (if any) from the server is: Peer certificate cannot be authenticated with given CA certificates
  *  External link https://gitforwindows.org/ failed: response code 0 means something's wrong.
             It's possible libcurl couldn't connect to the server or perhaps the request timed out.
             Sometimes, making too many requests at once also breaks things.
             Either way, the return message (if any) from the server is: SSL connect error
  *  External link https://oauth.net/2/ failed: response code 0 means something's wrong.
             It's possible libcurl couldn't connect to the server or perhaps the request timed out.
             Sometimes, making too many requests at once also breaks things.
             Either way, the return message (if any) from the server is: SSL connect error
```

response code 0 is not handled.
 Should I add something in the travis configuration? Add any configuration for HTMLProofer?

gjtorikian:
@fabiendv Try https://github.com/gjtorikian/html-proofer#ignoring-ssl-certificates and let me know if that works

fabiendv:
I got two errors left (https://gitforwindows.org/ and https://oauth.net). The first is gone.
Also, I added headers field with user-agent like below you sent me (just in case), and it doesn't work either.
Where this error comes from?

fabiendv:
@gjtorikian I have solved my issue [here](https://github.com/gjtorikian/html-proofer/issues/376#issuecomment-332767999). 
I was focused on status code 0 instead of SSL configuration.
Thanks for your answer

gjtorikian:
Great! Please let me know if you have any suggestions on improving visibility for that. I knew it was documented in an issue but I’d forgotten it was in the wiki, and I guess that isn’t immenintly searchable.

PR #327, opened 2016-04-15
Properly ignore alt without ignoring image

arvindsv:
* alt_ignore for an image does not ignore the whole image now. If it is missing, it will be reported.
* Also, this unearthed another issue, where srcset was not properly 'fixed' by prefixing it with http://.

Just so it is clear, without this change, any image that was alt_ignored-d, was ignored [here](https://github.com/gjtorikian/html-proofer/blob/0669699e8eb575456db97cfe6246591e37718e25/lib/html-proofer/check/images.rb#L21) in the checking process, and so, if the image it referred to is missing, the issue wouldn't be added to the final result.

### DIFF
--- a/lib/html-proofer/check/images.rb
+++ b/lib/html-proofer/check/images.rb
@@ -37,7 +37,7 @@ class ImageCheck < ::HTMLProofer::Check
         end
       end
 
-      if @img.alt.nil? || (empty_alt_tag? && !@img.ignore_empty_alt?)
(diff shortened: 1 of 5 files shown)


arvindsv:
By the way, srcset is not properly handled (as far as I can see). [srcset](http://w3c.github.io/html/semantics-embedded-content.html#element-attrdef-img-srcset) can actually have multiple strings, separated by a comma. Ideally, each should be separately checked. Currently, the assumption seems to be that srcset is just like src. So, I've left it like that, and have only corrected the prefixing to match href and src.

gjtorikian:
Thank you!

PR #343, opened 2016-06-26
Option to check the use of http scheme on img tags

maukoquiroga:
Although there's already support for enforcing the use of https in link tags, support for img tags is missing.

Extracted from https://github.com/sgmap/beta.gouv.fr/pull/253 and as a solution to https://github.com/sgmap/beta.gouv.fr/issues/74 (Firefox complaining of mixed-content when using non-HTTPS images).


Morendil:
Actually, don't strictly enforce `scheme == https`, but rather `scheme == nil || scheme == https`. Cases where the `src` of an `img` element is missing a scheme include relative URLs (which will not lead to mixed-content errors) and "absolute" URLs missing a scheme (which are common for including JS/CSS, and will not lead to mixed-content either).

This is a bit more work in the implementation: it requires checking the underlying `node` directly rather than checking the element constructed from the node, since the constructor for `HTMLProofer::Element` supplies `'https'` for `scheme` if the `node` doesn't have one.

maukoquiroga:
@Morendil I think the checking of the use of http scheme in the img src is well enough for the mixed-content scenario we're aiming to avoid. I've already updated the PR to that effect.

gjtorikian:
@maukoquiroga Thanks for your change! I've merged it into `master`.

I made one slight change (https://github.com/gjtorikian/html-proofer/commit/b39113ec80bd85691a1ed8aa6f62f74c4a718f2c) which is why I didn't straight merge this PR in. But your commits are still there. 👍

PR #350, opened 2016-08-10
Use regex for colon escape

Roang-zero1:
This replaces the detection of escaped colons to a regex, thereby voiding the need to escape backslashes.
Add test and fixture for an escaped colon.


gjtorikian:
Thank you! I am sorry for the delay in the merge, I have been mega busy lately.

PR #351, opened 2016-08-10
Add --internal-domains option

### LINKED ISSUE #349: Add setting for "internal" domains
I've come across the issue that I have multiple links to my own domain with a full path (e.g. https://www.example.com/page2) which are basically internal, but html-proofer treats them as external.

Currently there is the option to use url-swap (from the master branch) and swap with the "https://www,example.com":"" regex.

Unfortunately my site will be multilingual so I would have to add a regex for each protocol and each domain connected to my site.

I would propose a configuration option, which allows for a set of urls which will be treated as internal, thereby removing the leading protocol and domain.

For example:
`htmlproofer --home-domains www.example.com, de.example.com`
`<a href="https://www.example.com/about">About</a>` would check `"/about"`
`<a href="http://de.example.com/about">About</a>` would also check `"/about"`

Roang-zero1:
Adds an option to treat certain domains as internal.
Any domains specified will be removed from URLs via the url_swap feature.

Closes #349

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ The `HTMLProofer` constructor takes an optional hash of additional options:
 | `external_only` | Only checks problems with external references. | `false`
 | `file_ignore` | An array of Strings or RegExps containing file paths that are safe to ignore. | `[]` |
 | `http_status_ignore` | An array of numbers representing status codes to ignore. | `[]`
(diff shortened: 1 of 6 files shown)


gjtorikian:
Thank you!

PR #358, opened 2016-09-28
Add support for `base` element

### LINKED ISSUE #166: HTML base tag

gjtorikian:
From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base)

> The **HTML `<base>` element** specifies the base URL to use for all relative URLs contained within a document.

Successor to https://github.com/gjtorikian/html-proofer/pull/265, closes #166.

/cc @bkeepers

### DIFF
--- a/lib/html-proofer/check.rb
+++ b/lib/html-proofer/check.rb
@@ -1,7 +1,7 @@
 module HTMLProofer
   # Mostly handles issue management and collecting of external URLs.
   class Check
-    attr_reader :node, :element, :src, :path, :options, :issues, :external_urls
+    attr_reader :node, :html, :element, :src, :path, :options, :issues, :external_urls
 
     def initialize(src, path, html, options)
       @src    = src
--- a/lib/html-proofer/element.rb
+++ b/lib/html-proofer/element.rb
@@ -22,6 +22,8 @@ module HTMLProofer
       @type = check.class.name
       @line = obj.line
 
+      @html = check.html
+
       # fix up missing protocols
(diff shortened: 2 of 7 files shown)


lurch:
Bonus points for the branch-name :-)

lurch:
I don't know if it's ever used in practice, but according to the example in https://html.spec.whatwg.org/multipage/semantics.html#the-base-element 
```
In this example, a base element is used to set the document base URL:

<!DOCTYPE html>
<html lang="en">
    <head>
        <title>This is an example for the &lt;base&gt; element</title>
        <base href="http://www.example.com/news/index.html">
    </head>
    <body>
        <p>Visit the <a href="archives.html">archives</a>.</p>
    </body>
</html>
The link in the above example would be a link to "http://www.example.com/news/archives.html".
```
you can actually set base href to a page-name, and it looks like the 'effective' base href then becomes the directory that file is in.
Is it worth adding a unit-test for that?

bkeepers:
It's probably worth adding a test for it, but `Addressable::URI.join`, which is used in the implementation, does handle this scenario:

```ruby
>  Addressable::URI.join("http://www.example.com/news/index.html", "archives.html").to_s
=> "http://www.example.com/news/archives.html"
```

PR #361, opened 2016-10-20
Provide the Open Graph src URL by inheritance so internal-domains work

peternewman:


gjtorikian:
Perfect! Thank you! Released as 3.3.1.

peternewman:
Thanks. I'm not too hot on Ruby, but I assume what I did was reasonable?

gjtorikian:
Yep!

PR #362, opened 2016-11-01
HTML5 parsing and error checking

### LINKED ISSUE #318: Replace Nokogiri with something that validates HTML5
# Description

html-proofer **3.0.4** reports `Unexpected end tag : address` when there's an unordered list inside `<address>`. It does not report an error when an ordered list is used instead. I also don't see a reason why unordered lists shouldn't be valid inside `<address>`. 

[W3C specifies](https://www.w3.org/TR/html5/sections.html#the-address-element) the address element's content model as:

> Flow content, but with no heading content descendants, no sectioning content descendants, and no header, footer, or address element descendants.

_Flow content_ includes `<ul>` and `<ol>`.
# Steps to reproduce
### 1. Input

``` html
<!DOCTYPE HTML>
<html>
    <head></head>
    <body>
        <address>
            <ul></ul>
        </address>
    </body>
</html>
```
### 2. htmlproofer

``` bash
$ htmlproofer --check-html _site/test.html
Running ["ScriptCheck", "LinkCheck", "ImageCheck", "HtmlCheck"] on _site/test.html on *.html...


Checking 0 external links...
Ran on 1 file!


- _site/test.html
  *  Unexpected end tag : address (line 7)
htmlproofer 3.0.4 | Error:  HTML-Proofer found 1 failure!
```

jeremy:
Switches from Nokogiri to Nokogumbo. Depends on Nokogumbo >= 1.4.10 for [error reporting](https://github.com/rubys/nokogumbo/pull/46).

Introduces `--report-missing-doctype`, defaulting to false, to report on missing `<!DOCTYPE html>` at the beginning of the document.

Fixes #318

Pending:
- [x] spec updates for HTML5 parsing differences
- [x] whether to switch to HTML5 checking or introduce it as a new check
- [x] fix error line reporting (reported upstream at https://github.com/rubys/nokogumbo/issues/53)
- [x] pending Nokogumbo 2.0.0 release (including https://github.com/rubys/nokogumbo/pull/55)


gjtorikian:
Wow, I like a lot of this. I especially like the removal of the Nokogiri hack method--thank you for that! 

I'm a bit slammed at work at the moment so I won't be able to get this for a few days. For this task item:

> whether to switch to HTML5 checking or introduce it as a new check

I think we should just switch to HTML5 checking. I'd rather documents be modern and correct then continue to support separate HTML4/5 checks.

matkoniecz:
> I'm a bit slammed at work at the moment so I won't be able to get this for a few days.

ping?

Gemfile
```diff
@@ -1,3 +1,3 @@
-source 'http://rubygems.org'
+source 'https://rubygems.org'
```
gjtorikian: ❤️

html-proofer.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ['lib']
 
   gem.add_dependency 'mercenary',       '~> 0.3.2'
-  gem.add_dependency 'nokogiri',        '~> 1.5'
+  gem.add_dependency 'nokogumbo',       '~> 1.4', '>= 1.4.10'
```
gjtorikian: This seems awfully permissive. What's wrong with just `~> 1.4` ?

lib/html-proofer/check/html.rb
```diff
@@ -1,26 +1,29 @@
...
+    when SCRIPT_EMBEDS_MSG
+      options[:validation][:report_script_embeds]
+    when INVALID_TAG_MSG, INVALID_PREFIX
+      options[:validation][:report_invalid_tags]
+    when PARSE_ENTITY_REF
+      options[:validation][:report_missing_names]
+    when DOCTYPE_MSG
+      options[:validation][:report_missing_doctype]
+    else
+      true
```
gjtorikian: Super ❤️

spec/html-proofer/html_spec.rb
```diff
@@ -19,40 +19,40 @@
...
     proofer = run_proofer(html, :file, { :check_html => true })
-    expect(proofer.failed_tests.first).to match(/htmlParseEntityRef: expecting ';' \(line 2\)/)
+    expect(proofer.failed_tests).to eq []
   end
 
   it 'fails for mismatch between opening and ending tag' do
     html = "#{FIXTURES_DIR}/html/opening_and_ending_tag_mismatch.html"
     proofer = run_proofer(html, :file, { :check_html => true })
-    expect(proofer.failed_tests.first).to match(/Opening and ending tag mismatch: p and strong/)
+    expect(proofer.failed_tests.first).to include("@3:31: That tag isn't allowed here  Currently open tags: html, body, p, strong..\n<p>The quick <strong>brown fox</p>\n                              ^ (line 3)")
```
gjtorikian: Hmm. How do you feel about scrubbing the row/column position information in the front for all of these (`@3:31:`). It seems superfluous, given that we report the `line: ` at the end.

gjtorikian:
I like a lot of the clean-up here, thank you very much!

I am a bit concerned at some of the test changes, primarily where items were failing tests and have now flipped to not fail. Could you explain why that is? Is it because the HTML5 parsing is laxer than the previous HTML4 parsing?

html-proofer.gemspec
jeremy: This means *both* requirements apply. So it's `~> 1.4`, but at least `1.4.10`. That ensures that we can depend on the API from https://github.com/rubys/nokogumbo/pull/46.

spec/html-proofer/html_spec.rb
jeremy: 👍

jeremy:
> Could you explain why that is? Is it because the HTML5 parsing is laxer than the previous HTML4 parsing?

Precisely! Reflects where HTML5 rules differ from XHTML1/HTML4.

spec/html-proofer/links_spec.rb
```diff
@@ -64,10 +64,10 @@
     expect(proofer.failed_tests.first).to match(/internally linking to .\/notreal.html, which does not exist/)
   end
 
-  it 'fails for link with no href' do
+  it 'allows link with no href' do
     missingLinkHrefFilepath = "#{FIXTURES_DIR}/links/missingLinkHref.html"
     proofer = run_proofer(missingLinkHrefFilepath, :file)
-    expect(proofer.failed_tests.first).to match(/anchor has no href attribute/)
+    expect(proofer.failed_tests).to eq []
   end
```
jeremy: HTML5 allows links with no `href` attribute.

spec/html-proofer/links_spec.rb
```diff
@@ -208,16 +208,16 @@
...
 
-  it 'fails for absent href within link elements' do
+  it 'allows missing href on link elements' do
     head_link = "#{FIXTURES_DIR}/links/head_link_href_absent.html"
     proofer = run_proofer(head_link, :file)
-    expect(proofer.failed_tests.first).to match(/anchor has no href attribute/)
+    expect(proofer.failed_tests).to eq []
   end
```
jeremy: Ditto. HTML5 allows links with no `href` attribute.

gjtorikian:
@jeremy Would you mind merging master to fix the merge conflict? After that I think this should be good to go. Thank you!

jeremy:
@gjtorikian Rebased.

gjtorikian:
I completely forgot about this! Sorry!

gjtorikian:
Unfortunately I still can't accept this just yet. `Nokogiri::HTML5` does not appear to preserve the line numbers correctly. I opened https://github.com/rubys/nokogumbo/issues/53 and am blocked on that.

jeremy:
Nokogumbo doesn't include a gemspec, so we can't bundle it from git, and the fix isn't included in the recent 2.0.0.pre.alpha release. That means we can't depend on a fixed version to test this PR, so I branched nokogumbo (https://github.com/rubys/nokogumbo/pull/104) to provide a non-Rakefile-embedded gemspec for temporary use.

Gemfile
```diff
@@ -1,3 +1,5 @@
...
+
+gem 'nokogumbo', git: 'https://github.com/jeremy/nokogumbo', branch: 'gemspec'
```
jeremy: Until Nokogumbo includes a gemspec (https://github.com/rubys/nokogumbo/pull/104) or releases a version with line number support.

html-proofer.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ['lib']
 
   gem.add_dependency 'mercenary',       '~> 0.3.2'
-  gem.add_dependency 'nokogiri',        '~> 1.8.1'
+  gem.add_dependency 'nokogumbo',       '>= 2.0.0.alpha', '< 3'
```
jeremy: Pending 2.0.0 release.

jeremy:
Note that a bunch of line numbers are still incorrect `(line 0)`, so the upstream fix is likely insufficient.

Another route is to support what we can and defer to the underlying library on specifics, allowing users to parse HTML5 today despite the incomplete support.

jeremy:
Switched nokogumbo dep to track git master pending 2.0.0 release.

Gemfile
```diff
@@ -1,3 +1,5 @@
-source 'http://rubygems.org'
+source 'https://rubygems.org'
+
+gem 'nokogumbo', git: 'https://github.com/rubys/nokogumbo'
```
jeremy: Until Nokogumbo 2.0.0 is released.

gjtorikian: Let's lock this to [a `:ref` as well](https://bundler.io/v1.12/git.html) to avoid a situation where `master` might update and break things.

jeremy: Note this is the project's own Gemfile, not a runtime gem dependency, and this PR can't be merged until 2.0.0 is released, at which point the Gemfile entry will be removed. So we aren't exposing ourselves or any users to possible breakage by tracking the git branch to demonstrate the CI passes. If anything, we're avoiding surprises from subsequent prerelease updates that *would* break CI 😊

stevecheckoway:
> Note that a bunch of line numbers are still incorrect `(line 0)`, so the upstream fix is likely insufficient.
> 
> Another route is to support what we can and defer to the underlying library on specifics, allowing users to parse HTML5 today despite the incomplete support.

Do you have examples of where the line numbers are going wrong? I can take a look. One potential issue is that libxml2 only supports line numbers up to 65535. (There's a build-time option that increases that to a 4-byte integer, but only for some nodes—text nodes, if memory serves—and other nodes have to hunt around to find nearby text nodes to guess what line they're on. I didn't take advantage of that.)

jeremy:
@stevecheckoway Can see the failing specs on this PR (https://travis-ci.org/gjtorikian/html-proofer/jobs/424573307#L872) or reproduce by running the suite locally.

stevecheckoway:
If Nokogiri is using the system library (as is the case here) and the nokogumbo build process can’t find the libxml2 headers, then line numbers aren’t supported. I suspect that’s what’s happening here but I can’t tell from the log.

fulldecent:
Both blockers above are now fixed upstream. Can this proceed?

fulldecent:
Who needs resources to get this done?

stevecheckoway:
@fulldecent, when you say both blockers are fixed upstream, what are you referring to? As far as I can tell, Nokogiri doesn't have an API for setting the line number, just an API for [getting a line number](https://github.com/sparklemotion/nokogiri/blob/c021f65d94f979bb8656d989949fc1bfc57b91a6/ext/nokogiri/xml_node.c#L1327).

Absent that, Nokogumbo is [unable to set line numbers](https://github.com/rubys/nokogumbo/blob/master/ext/nokogumbo/nokogumbo.c#L320) if it is forced to use the Ruby API (which happens when [this condition](https://github.com/rubys/nokogumbo/blob/master/ext/nokogumbo/extconf.rb#L101) is false).

stevecheckoway:
I filed a PR with Nokogiri to add a line setting API (referenced above) and I added https://github.com/rubys/nokogumbo/issues/122 to track the issue in Nokogumbo.

fulldecent:
Thank you for providing the update. May https://github.com/rubys/nokogumbo/issues/122 please be referenced as a blocker at top?

fulldecent:
We have a path forward with https://github.com/rubys/nokogumbo/pull/130

spec/html-proofer/html_spec.rb
fulldecent: This is fixed at https://github.com/gjtorikian/html-proofer/pull/362/files#diff-9549a780abfd37b3a235e0289440f6d6R40

fulldecent:
In Nokogumbo we are waiting for https://github.com/rubys/nokogumbo/pull/130 so that we can use it here.

fulldecent:
Fixed merge conflicts. And still have a few failing tests.

fulldecent:
Requesting help here to fix unit tests and review this PR

gjtorikian:
For some reason I cannot push to this PR? But here is the diff necessary to fix the tests:

```diff
diff --git a/lib/html-proofer/middleware.rb b/lib/html-proofer/middleware.rb
index b1e7eab..a154261 100644
--- a/lib/html-proofer/middleware.rb
+++ b/lib/html-proofer/middleware.rb
@@ -67,7 +67,7 @@ module HTMLProofer
           'response',
           Middleware.options
         ).check_parsed(
-          Nokogiri::HTML(Utils.clean_content(html)), 'response'
+          Nokogiri::HTML5(html), 'response'
         )
 
         if parsed[:failures].length > 0
```

fulldecent:
@gjtorikian I was able to put that in here https://github.com/jeremy/html-proofer/commit/35f85cb4ddbe8616ed968d7a8de42a10b2ab7d92

gjtorikian:
@fulldecent Do you have push access to his repo? Normally I can make commits directly to PRs but I can't seem to manage it with this one. Looks like there's just one failure left.

fulldecent:
@gjtorikian Yes, but I can only do it with the web interface. Here is the link I'm using: 

https://github.com/jeremy/html-proofer/tree/html5

fulldecent:
@gjtorikian does this work for you now?

gjtorikian:
I do believe so, but I haven't had a chance to look into it a fix as yet.

gjtorikian:
I believe that should do it. I'll take a look at the merge conflicts as soon as I am able.

gjtorikian:
Thanks everyone. This is looking ready to go. I will merge, release, and write release notes as soon as possible--I only just had a quick ten minutes right now to do all the merge conflict fixes.

jeremy:
Thanks for taking this to the finish line @gjtorikian!! 🎉

Issue #378, opened 2017-02-06
--url-ignore matching


ilg-ul:
In one of my sites, the proofer reported:

```
*  External link https://jekyllrb.com/ failed: response code 0 means something's wrong.
```

Initially I added `--url-ignore "https://jekyllrb.com/"`, but this was not useful.

Inspecting the input file, revealed the reference has no trailing `/`.

```
<p><a href="https://jekyllrb.com">Jekyll</a> is a simple, blog-aware, static site generator.</p>
```

Changing the command line to `--url-ignore "https://jekyllrb.com"`, silenced the error.

So, the problem is an inconsistency between the input file (without slash), the error message (without slash) and the option required to silence the error.

Suggestions:

- be sure the URL displayed in the error message is exactly the one that needs t be passed to `--url-ignore`
- if the `--url-ignore` path does not end in a slash, also try to match the one with a slash

gjtorikian:
I like it! I've assigned this for future work.

PR #393, opened 2017-03-04
Ensure lack of slash in error messages

gjtorikian:
Typheous unfortunately occasionally adds an extra slash when it attempts
to check external URLs. This change at least removes that slash from
the URL to make future grepping easier, if not 100% accurate.

Closes https://github.com/gjtorikian/html-proofer/issues/378.


fulldecent:
Can this please be reported upstream to Typhoeus?

As current, valid URLs ending with a `/` are misreported by html-proofer.

Issue #369, opened 2016-12-15
Configurable HTMLProofer::Utils::STORAGE_DIR?


tisba:
It would be great, if you could specify the `HTMLProofer::Utils::STORAGE_DIR`. On our CI we fetch (when available) the cache manifest and copy that into place. It would be nice if you could overwrite this w/o monkey patching this constant.

Is this something you would accept a PR for?

gjtorikian:
Yeah, that sounds totally fine to me. Bring the PR and I'll take a look!

gjtorikian:
Did you want to open a PR for this?

tisba:
Thanks for the reminder, I've opened up a PR.

PR #395, opened 2017-03-04
Make cache storage directory configurable

tisba:
This PR will add a `--storage-dir` option where you can configure the directory where the cache log file will be read from and written to.

I took the liberty to update the cache specs a bit to be more idiomatic RSpec/Ruby. Most of the change was required because the cache log location is no longer a constant nor can it be injected via stubbing a constant.

Fixes https://github.com/gjtorikian/html-proofer/issues/369


gjtorikian:
Great, thank you!

PR #396, opened 2017-03-06
Update README.md to reflect correct cache.log path

sjauld:
README.md states that the log file is called *tmp/.htmlproofer*. It should be *tmp/.htmlproofer/cache.log*.


gjtorikian:
Thank you!

PR #399, opened 2017-03-07
Remove cache from previous tests

mattclegg:
I'm not sure why these are added to the repository but I've added a .gitignore file to the folder `spec/html-proofer/fixtures/vcr_cassettes` to prevent them being added again.


gjtorikian:
I followed the conventions of other repositories that use cassettes, like https://github.com/octokit/octokit.rb/tree/d9b2ed79a0465e34ae03baadf05b5173df2c200e/spec/cassettes.

The idea is that I want to play the HTTP interaction _once_, and don't care about it after that. I only want to act on the responses. By removing the cache, Travis CI is going to have to redo the network requests each and every time. A push is made.

Have you seen suggestions otherwise to remove/ignore this dir?

PR #401, opened 2017-03-15
Escape quotes in anchors

henri-tremblay:
Having an id containing a quote is not allowed in HTML5. Having a quote currently crashes the link checking. This fix should solve that.

### DIFF
--- a/lib/html-proofer/check/links.rb
+++ b/lib/html-proofer/check/links.rb
@@ -107,10 +107,12 @@ class LinkCheck < ::HTMLProofer::Check
 
   def hash_check(html, href_hash)
     decoded_href_hash = URI.decode(href_hash)
-    html.xpath("//*[case_insensitive_equals(@id, '#{href_hash}')]", \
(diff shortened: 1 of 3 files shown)


gjtorikian:
Bummer. Can you add a test to check this as well? Thank you!

henri-tremblay:
I'm not really a ruby expert. Nokogiri is as usual giving me a hard time. So I failed to run your tests and don't really want to loose time at it.

A test would probably look like this:

In `links_spec.rb`:
```ruby
  it 'works with quotes in the hash href' do
    hash_href = "#{FIXTURES_DIR}/links/quote.html"
    proofer = run_proofer(hash_href, :file, { :allow_hash_href => true })
    expect(proofer.failed_tests.length).to eq 0
  end
```

In a new `quote.html` file:
```html
<html>
<body>
<a href="#let's-go"></a>
</body>
</html>
```

henri-tremblay:
Ok. I finally found the trick. So I've pushed a test.

gjtorikian:
Thanks!

PR #402, opened 2017-03-17
Yell documentation is not helpful, inline needed information

fulldecent:


fulldecent:
I am not responsible for breaking the build.

gjtorikian:
Thank you!

PR #407, opened 2017-03-17
Set Typhoeus connection timeout, fixes #231, fixes #340

### LINKED ISSUE #231: External Link Checking - Timeouts
Hi folks,

I'm using `html-proofer` to test a website automatically on [Travis CI](https://travis-ci.org/).
Unfortunately, the builds are timing out because the external link checking process is taking to long.
Here's an example:
https://travis-ci.org/squeak-smalltalk/squeak.org/builds/72423569#L236

There are two possible solutions that I can think of:
1. Display progress information (Travis builds fail if there's no output being generated in 10min)
2. Add a `--timeout` option. If a web requests takes longer, print a warning but continue.

Feedback?

Best,
Fabio

### LINKED ISSUE #340: External links pointing to non-responsive servers cause stall
Checking an external link resulting in `Received a 0 for http://website/ in ./_site/page.html` caused a 10 minute stall in html-proofer.

Webserver is un-pingable, HTTP requests hang. Can html-proofer implement a timeout for these checks?

fulldecent:
If you can't even connect to the server in 10 seconds (5 internet-years), then fail that connection.

This should solve many people's problems with `htmlproofer` crashing their Travis builds.

The default are more than reasonable. And of course people can override the defaults by using the Ruby API.

### DIFF
--- a/lib/html-proofer/configuration.rb
+++ b/lib/html-proofer/configuration.rb
@@ -34,7 +34,8 @@ module HTMLProofer
       :headers => {
         'User-Agent' => "Mozilla/5.0 (compatible; HTML Proofer/#{HTMLProofer::VERSION}; +https://github.com/gjtorikian/html-proofer)"
       },
-      :connecttimeout => 10
+      :connecttimeout => 10,
+      :timeout => 30
     }
 
     HYDRA_DEFAULTS = {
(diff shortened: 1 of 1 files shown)


fulldecent:
Perhaps we could also set

    :timeout 60

This is a very generous default option that nobody should complain about. Maybe that should be a battle for a different day.

gjtorikian:
> 5 internet-years

😆 

Thank you!

PR #413, opened 2017-03-28
Check for SRI/CORS in CSS files

### LINKED ISSUE #406: Check for SRI/CORS in CSS files
html-proofer currently checks for SRI/CORS attributes in Javascript links.

It should also do this for stylesheet links.

This is NOT a new command-line option. This is correcting the behavior of the current `--check-sri` option to match behavior as documented.

---

We can duplicate work for scripts here: https://github.com/gjtorikian/html-proofer/pull/388/files

Here is where scripts are checked: https://github.com/ilyalyo/html-proofer/blob/0cc1e95fcb86e505ad27614299f184aade44e5a2/lib/html-proofer/check/scripts.rb

Here is where we want links to be checked: https://github.com/ilyalyo/html-proofer/blob/0cc1e95fcb86e505ad27614299f184aade44e5a2/lib/html-proofer/check/links.rb

ilyalyo:
close #406

### DIFF
--- a/lib/html-proofer/check/links.rb
+++ b/lib/html-proofer/check/links.rb
@@ -40,6 +40,7 @@ class LinkCheck < ::HTMLProofer::Check
       next if @link.non_http_remote?
 
       if !@link.internal? && @link.remote?
+        check_sri(line, content) if @link.check_sri?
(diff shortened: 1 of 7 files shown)


gjtorikian:
Thank you!

PR #420, opened 2017-04-14
handle base without href

mlinksva:
[`<base>`](https://html.spec.whatwg.org/multipage/semantics.html#the-base-element) can have `href` and/or `target` attributes. Currently, `href` is not present obtains:

> htmlproofer 3.6.0 | Error:  Can't convert NilClass into String.

This PR allows `<base>` to not have an `href` attribute.


gjtorikian:
Thank you!

Issue #118, opened 2014-09-28
Ignore <a href="#"> when checking internal links


akoeplinger:
Using such anchors is quite a common practice (e.g. by [Bootstrap Dropdowns](http://getbootstrap.com/components/#btn-dropdowns)) and generates the following error:

> index.html: linking to internal hash # that does not exist

I think they shouldn't trigger an error as they are just used as a placeholder and not for linking to a specific part of the document.

ssbarnea:
I think that this bug should be reopen, Jekyll should ignore this one by default, because as already pointed, that's valid, and also quite common. Default settings should be friendly.

penibelst:
> Jekyll should ignore this one by default.

You mean Proofer, right?

jacebrowning:
I think `htmlproof <site> --href-ignore "#"` should be the default.

einverne:
It seems in version 3.0.5 there is no option `--href-ignore "#"`. However, after I add `--allow-hash-href` or `--url-ignore "#"`. I still got error like this:

    linking to internal hash #left that does not exist (line 205)

einverne:
After I lot of search, I figure I have to write RegExps like `--url-ignore "/#.*/` to ignore "#left" etc internal hash.

fulldecent:
I'm just writing a note here to clarify since I see other people are linking to this page. To set the record straight: 

> If you use `<a href="#xxx">` when `#xxx` does not exist, then you are doing it wrong.

A lot of people do it, but it's wrong. Usually the correct solution is to have `<span class=...>` or `<div class=...>`. The Bootstrap people are pretty smart. Their examples are full of `"#"` but those represent links for content you still need to make.

*Exception*: if you have something like Gmail where the fragment is `#inbox-message-12546` and it intentionally changes to reflect page state then maybe that is special. But in that case, I would argue that you should have `<body id="inbox-message-12546">`.

aduth:
While true that empty anchor fragment `#` is often misused, it is considered in the HTML specification as a valid value referencing the top of the current document:

>If fragment is the empty string, then the indicated part of the document is the top of the document; return.

https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-the-fragment-identifier

It could be used this way for "Back to Top" links shown at the bottom of a page, for example.

See also #421, which added explicit support for `#top` links (which pass currently using `htmlproofer`).

fulldecent:
[...]

PR #421, opened 2017-04-25
support links to the #top of pages

afeld:
Per the HTML5 spec:

> If `fragid` is an ASCII case-insensitive match for the string `top`, then the indicated part of the document is the top of the document

https://www.w3.org/TR/html5/single-page.html#scroll-to-fragid

/cc #118

### DIFF
--- a/lib/html-proofer/check/links.rb
+++ b/lib/html-proofer/check/links.rb
@@ -107,13 +107,20 @@ class LinkCheck < ::HTMLProofer::Check
 
   def hash_check(html, href_hash)
     decoded_href_hash = URI.decode(href_hash)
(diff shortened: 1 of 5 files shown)


gjtorikian:
This is perfect. Thank you!

aduth:
A few lines prior in the specification also notes the same for empty fragments `href="#"`:

>If fragment is the empty string, then the indicated part of the document is the top of the document; return.

It's especially relevant for the related issue #118. I made a related comment at https://github.com/gjtorikian/html-proofer/issues/118#issuecomment-653046463. Since empty fragments are often _misused_, it might still make sense to err on the side of flagging it as an error, and encouraging people to use `#top` in case they explicitly intend "back to top".

Issue #408, opened 2017-03-19
Can I configure parallel from the command line?


pauldambra:
I asked this question [on stackoverflow](http://stackoverflow.com/questions/42892839/can-i-configure-parallel-when-calling-html-proofer-from-a-script) originally....

I'm calling [html proofer][1] from a script like so

    #! /bin/bash
    
    set -eu
    
    bundle exec htmlproofer \
      _site \
      --file-ignore /amp/,/.git/ \
      --check-favicon \
      --check-html \
      --check-opengraph

I've added the [parallel gem][2] to my Gemfile and that makes no difference to run time.

Should it have done? Can I hook in parallel in the script?

Thanks!

  [1]: https://github.com/gjtorikian/html-proofer
  [2]: https://github.com/grosser/parallel

pauldambra:
Ah, just forked and running locally. So, I can see that I can't already.

Also, that there's a single tags.html page that takes almost all of the time in the run because, I assume, it has 1177 internal links.

gjtorikian:
Ugh, this is an issue also related to https://github.com/gjtorikian/html-proofer/issues/379.
 
I really need to make this happen.

pauldambra:
No need to apologise! HTML proofer is awesome and there are always more things to build than time to build them :)

fulldecent:
We currently do not have command line support for parallel configuration.

BUT I have fully documented how to configure this "advanced" feature using Ruby. And there are copy/pastable usage guides for Travis.

> https://github.com/gjtorikian/html-proofer/wiki/Using-HTMLProofer-From-Ruby-and-Travis

fulldecent:
@pauldambra Could you please confirm if wiki is helpful enough for you to make it work the way you like?

pauldambra:
I'm calling the CLI in a bash script. I could call out to a ruby script from there so can probably use that... Shall I try that (can't be today unfortunately) and then I can PR against the wiki page of I can think of any improvement for my use-case?

fulldecent:
@pauldambra yes please, this would be most helpful.

Technical note, you are able to directly edit the wiki, no PR required.

tiagoboldt:
My 2 cents, for the sake of keeping usage simple and clutter free, it would be very much useful if parallel could be used from the command line.

pauldambra:
I've just converted my bash script to a ruby script. So...

Issue #379, opened 2017-02-15
Allow Typheous options to be set from `bin`/command line


gjtorikian:

fulldecent:
This will get ugly fast. I can't think of a clean way to specify arbitrary Ruby on the command line.

Here is another issue that can be solved with Typheous configuration: https://github.com/gjtorikian/html-proofer/issues/86

I recommend to create turn on wiki pages and create a page about Typheous hacks that people can use. This will require people to use Ruby but they can get what they need done and it does not clog up this project with implement everyone's favorite Typheous feature.

Also long term, if we take that approach then in the future we could look at the activity on that blog and make a better decision on implementing the command-link Typheous support discussed here (and reopen this issue).

fulldecent:
* A 100% usable workaround is in the wiki
* In 6 months, nobody else has contributed to that wiki (there's a few TODOs outstanding)
* Nobody submitted a pull request
* It is questionable if this issue promotes project scope

**Recommendation: close issue**

Perhaps we can reopen the issue after somebody else cares to update the wiki into a rock solid page (address all TODO items). But for now it seems this is going nowhere.

Issue #194, opened 2015-03-04
command line ssl_verifypeer false


fvanderbiest:
I'm currently proofing my jekyll-generated site with `htmlproof ./_site` and I'd like to get rid of those warnings:

```
External link failed: 302 Peer certificate cannot be authenticated with given CA certificates
External link  failed: 0 SSL connect error
External link  failed: 302 SSL connect error
```

I could not find any documentation on how to disable ssl_verifypeer using the htmlproof binary.

penibelst:
The usage of the option `ssl_verifypeer` is a mystery. See #141.

penibelst:
Can you try to curl your link and show us the output?

```
curl --url https://www.example.com/ --verbose
```

fvanderbiest:
Sure.
```
$ curl --url https://sdi.georchestra.org/ci/ --verbose
* Hostname was NOT found in DNS cache
*   Trying 195.154.179.143...
* Connected to sdi.georchestra.org (195.154.179.143) port 443 (#0)
* successfully set certificate verify locations:
*   CAfile: none
  CApath: /etc/ssl/certs
* SSLv3, TLS handshake, Client hello (1):
* SSLv3, TLS handshake, Server hello (2):
* SSLv3, TLS handshake, CERT (11):
* SSLv3, TLS handshake, Server key exchange (12):
* SSLv3, TLS handshake, Server finished (14):
* SSLv3, TLS handshake, Client key exchange (16):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSL connection using ECDHE-RSA-AES256-GCM-SHA384
* Server certificate:
* 	 subject: C=FR; CN=sdi.georchestra.org; emailAddress=webmaster@georchestra.org
* 	 start date: 2014-12-30 15:31:45 GMT
* 	 expire date: 2015-12-31 19:14:52 GMT
* 	 subjectAltName: sdi.georchestra.org matched
* 	 issuer: C=IL; O=StartCom Ltd.; OU=Secure Digital Certificate Signing; CN=StartCom Class 1 Primary Intermediate Server CA
* 	 SSL certificate verify ok.
```

An other one
```
$ curl --url https://gitlab.geo.gob.bo/adsib/georchestra_ansible/tree/master --verbose
* Hostname was NOT found in DNS cache
[...]

PR #433, opened 2017-05-18
Document difference between command line and Ruby implementation

fulldecent:
This change to the project scope clarifies that "advanced" usage of this project will require using Ruby.

It would allow us to close issues as out of scope:

* #408 
* #379 
* #194 

Also it will allow us to accept a change to the README which would simplify the recommended use case (command line) while allowing further, advanced discussion in a separate file or wiki like https://github.com/gjtorikian/html-proofer/wiki/Using-HTMLProofer-From-Ruby-and-Travis


gjtorikian:
Thank you!

PR #437, opened 2017-06-24
add STAT output option

### LINKED ISSUE #387: Standardize output format, use STAT format
**BLOCKER: https://github.com/fulldecent/structured-acceptance-test/issues/44**

Output of html-proofer is not designed to be machine readable. There is a simple output format that is machine readable which may be useful to consider.

https://github.com/fulldecent/structured-acceptance-test

This standard includes a machine-readable json output format which can be parsed by other tools and rendered in realtime in Atom and other editors. Additionally, that project includes Ruby classes for describing findings and outputting them. Out of the box it supports pretty output formats (with color) and JSON output formats.

https://github.com/fulldecent/structured-acceptance-test/tree/master/implementations/ruby

# Work scope

- [ ] Clearly define the output format of html-proofer findings (i.e. errors and warnings)
- [ ] Implement standardized StatModule as the in-memory format for findings
- [ ] Implement default output format matching existing html-proofer output from
- [ ] Create new option `--stat` which follows stat standardized output format

# Future scope

 - [ ] Propose and implement new and prettier standard output format

ilyalyo:
Closes #387

### DIFF
--- a/lib/html-proofer/check.rb
+++ b/lib/html-proofer/check.rb
@@ -1,13 +1,14 @@
 module HTMLProofer
   # Mostly handles issue management and collecting of external URLs.
   class Check
-    attr_reader :node, :html, :element, :src, :path, :options, :issues, :external_urls
+    attr_reader :node, :html, :element, :src, :path, :options, :external_urls, :stat
 
-    def initialize(src, path, html, options)
+    def initialize(src, path, html, options, stat)
       @src    = src
       @path   = path
       @html   = remove_ignored(html)
       @options = options
+      @stat   = stat
       @issues = []
       @external_urls = {}
     end
@@ -23,6 +24,8 @@ module HTMLProofer
 
     def add_issue(desc, line: nil, status: -1, content: nil)
       @issues << Issue.new(@path, desc, "#{self::class } failed", line: line, status: status, content: content)
+      @stat.findings.push(Issue.new(@path, desc, "#{self::class } failed", line: line, status: status, content: content))
+      @stat.print_finding if @options[:stat]
     end
 
     def add_to_external_urls(url)
--- a/lib/html-proofer/check.rb
+++ b/lib/html-proofer/check.rb
@@ -3,7 +3,7 @@ module HTMLProofer
   class Check
     attr_reader :node, :html, :element, :src, :path, :options, :external_urls, :stat
 
-    def initialize(src, path, html, options, stat)
+    def initialize(stat, src, path, html, options)
       @src    = src
       @path   = path
       @html   = remove_ignored(html)
--- a/lib/html-proofer/check.rb
+++ b/lib/html-proofer/check.rb
@@ -22,7 +22,7 @@ module HTMLProofer
     end
 
     def add_issue(desc, line: nil, status: -1, content: nil)
-      @issues << Issue.new(@path, desc, line: line, status: status, content: content)
+      @issues << Issue.new(@path, desc, "#{self::class } failed", line: line, status: status, content: content)
     end
 
     def add_to_external_urls(url)
--- a/lib/html-proofer/log.rb
+++ b/lib/html-proofer/log.rb
@@ -5,16 +5,23 @@ module HTMLProofer
   class Log
     include Yell::Loggable
 
-    def initialize(log_level)
-      @logger = Yell.new(format: false, \
-                         name: 'HTMLProofer', \
-                         level: "gte.#{log_level}") do |l|
-        l.adapter :stdout, level: %i[debug info warn]
-        l.adapter :stderr, level: %i[error fatal]
+
+    def initialize(log_level, hide)
+      if !hide
+        @logger = Yell.new(format: false, \
+                           name: 'HTMLProofer', \
+                           level: "gte.#{log_level}") do |l|
+          l.adapter :stdout, level: %i[debug info warn]
+          l.adapter :stderr, level: %i[error fatal]
+        end
+      else
+        @hide = true
       end
     end
 
     def log(level, message)
+      return if @hide
+
       color = case level
               when :debug
                 :light_blue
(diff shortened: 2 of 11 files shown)


fulldecent:
A few items here:

- There are unnecessary change in white space
- Maintainer should be Garen Torikian
- Is it possible to remove the `HTMLProofer::Issue` and instead use instances of `StatModule::Finding`? The `def initialize(path, desc, rule, line: nil, status: -1, content: nil)` initializer is convenient. Maybe HTML Proofer should extend or subclass `StatModule::Finding` to add this convenience initializer.
- This PR sends all issues to STAT at once. Instead it should stream them as identified.

ilyalyo:
Updated, I left class Issue and make it extends Finding.
Also now it streams findings if stat option is specified.

lib/html-proofer/check.rb
```diff
@@ -22,7 +23,9 @@ def run
     end
 
     def add_issue(desc, line: nil, status: -1, content: nil)
-      @issues << Issue.new(@path, desc, line: line, status: status, content: content)
+      @issues << Issue.new(@path, desc, "#{self::class } failed", line: line, status: status, content: content)
+      @stat.findings.push(Issue.new(@path, desc, "#{self::class } failed", line: line, status: status, content: content))
+      @stat.print_finding if @options[:stat]
```
gjtorikian: Doesn't this mean that on each issue, a new statement will be printed? Wouldn't it be preferable to provide that information at the end of all the runs? In other words, instead of iterating and printing `@issues`, it seems to make more sense to do something like:

```
@issues.each do |issue|
  @stat.print(issue)
```

Where `@stat.print` massages the `issue` format to whatever it likes.

lib/html-proofer/check.rb
```diff
@@ -1,13 +1,14 @@
 module HTMLProofer
   # Mostly handles issue management and collecting of external URLs.
   class Check
-    attr_reader :node, :html, :element, :src, :path, :options, :issues, :external_urls
+    attr_reader :node, :html, :element, :src, :path, :options, :external_urls, :stat
 
-    def initialize(src, path, html, options)
+    def initialize(src, path, html, options, stat)
```
gjtorikian: Please put `stat` before `options`, as this is standard Ruby formatting.

lib/html-proofer/log.rb
```diff
@@ -5,16 +5,23 @@ module HTMLProofer
...
 
-    def initialize(log_level)
-      @logger = Yell.new(format: false, \
-                         name: 'HTMLProofer', \
-                         level: "gte.#{log_level}") do |l|
-        l.adapter :stdout, level: %i[debug info warn]
-        l.adapter :stderr, level: %i[error fatal]
+
+    def initialize(log_level, hide)
+      if !hide
```
gjtorikian: Instead of this, please do

```
 @logger = HTMLProofer::Log.new(@options[:log_level]) unless @options[:stat]
``` 

earlier in the code. Every method in this class having to do `return if` seems like a smell.

lib/html-proofer/runner.rb
```diff
@@ -27,11 +27,14 @@ def initialize(src, opts = {})
...
-      @failures = []
     end
 
     def run
+      @stat = init_stat
+      if @options[:stat]
+        @stat.print_header
+      end
+
       @logger.log :info, "Running #{checks} on #{@src} on *#{@options[:extension]}... \n\n"
```
gjtorikian: Since we removed `return if @hide` above, this area seems like a good place to refactor:

``` ruby
if @options[:stat]
  @state = init_stat
  @stat.print_header
else
  @logger.log :info, "Running ..."
end
```

and so on. Basically flip on the logging iff `@stat` is `false`.

lib/html-proofer/utils.rb
```diff
@@ -34,5 +35,17 @@ def clean_content(string)
       end
     end
     module_function :clean_content
+
+
+    def init_stat
```
gjtorikian: Delete this extraneous space.

lib/html-proofer/utils.rb
```diff
@@ -34,5 +35,17 @@ def clean_content(string)
...
+      process = StatModule::Process.new('HTMLProofer')
+      process.version = "#{HTMLProofer::VERSION}"
+      process.description = 'If you generate HTML files, then this tool might be for you.'
+      process.maintainer = 'Garen Torikian'
+      process.email = 'gjtorikian@gmail.com'
+      process.website = 'https://github.com/gjtorikian/html-proofer'
```
gjtorikian: Please load these from the Gemfile itself, so that constants do not need to be recorded in two places.

fulldecent:
@ilyalyo Please follow up with these requested changes.

lib/html-proofer/check.rb
ilyalyo: We print new statement immediately only if stat option is checked, it is requirement of stat:

> INCREMENTAL OUTPUT: Each finding SHOULD be delimited by a new line. A computer program with STAT-compliant output SHOULD immediately output each finding as available (e.g. use flush()). This allows the consumer to begin processing the results right away.
https://github.com/fulldecent/structured-acceptance-test/blob/master/Stat-Output.md

lib/html-proofer/check.rb
ilyalyo: Got you, fixed.

lib/html-proofer/log.rb
ilyalyo: I updated PR with this solution, but it also has its minuses - we have to be sure that we use only one instance of logger class, one that we created at Runner class and we have to check that @logger is not null every time we use it.

lib/html-proofer/utils.rb
ilyalyo: fixed

lib/html-proofer/utils.rb
ilyalyo: fixed

ilyalyo:
@gjtorikian ready to review.

PR #445, opened 2017-08-22
Send a sensible Accept header when testing external links

timrogers:
When testing external links, we currently send a `User-Agent` but no `Accept` header, which specifies the kind of content we want to get back.

This causes problems with testing some URLs, since they return a 404 (or other 4XX error code) as they don't know what format to return in.

This adds a default value, using the header [sent by Chrome and Safari](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values) as a sensible default.

### DIFF
--- a/lib/html-proofer/configuration.rb
+++ b/lib/html-proofer/configuration.rb
@@ -32,7 +32,8 @@ module HTMLProofer
     TYPHOEUS_DEFAULTS = {
       followlocation: true,
       headers: {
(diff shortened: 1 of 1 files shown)


gjtorikian:
Cool, thanks!

timrogers:
I'd be interested in becoming a maintainer I think, by the way 😄

PR #451, opened 2017-09-14
Change to use colorize gem

### LINKED ISSUE #449: Replaced 'colored' with 'colorize'
https://github.com/defunkt/colored is unmaintained, and not updated for 8 years now. https://github.com/fazibear/colorize is an alternative that's being looked after.

Floppy:
Colored is no longer maintained. Resolves #449, and will prevent a bunch of outdated dependency warnings for people using html-proofer.

There currently aren't any tests for the `Log` class, as far as I can see. Would you want some added in order to merge this?

### DIFF
--- a/html-proofer.gemspec
+++ b/html-proofer.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency 'mercenary',       '~> 0.3.2'
   gem.add_dependency 'nokogiri',        '~> 1.7'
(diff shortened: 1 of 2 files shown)


gjtorikian:
Nah. Thanks for the change!

Floppy:
And just like that, html-proofer no longer makes my [DependencyCI](https://dependencyci.com/github/SomethingNewUK/redirector/builds/39) builds fail! I'm so happy right now 😁

PR #452, opened 2017-09-14
More robust redirect detection

Floppy:
An improvement on #448 that specifically checks for a zero-time refresh meta tag.


gjtorikian:
Way better. Thank you!

Issue #409, opened 2017-03-21
Error:  string contains null byte


ilyalyo:
Link like:
```html
<a href="https://somelink.com/content.swf#">link</a>
```
with empty bookmark produce 'htmlproofer 3.5.0 | Error:  string contains null byte' error
htmlproofer runs with such options:
```
bundle exec htmlproofer --check-sri --check-external-hash --check-html --check-img-http --check-opengraph --enforce-https --timeframe 6w ./_site/index.html
```
Operating System Details:
Ubuntu 12.04.5 LTS

fulldecent:
I cannot reproduce. Here is what I tried:

```sh
mkdir tmp 
cd tmp
echo -e "source \"https://rubygems.org\"\ngem \"html-proofer\"" > Gemfile
bundle install
echo '<a href="index.html#">link</a>' > index.html
bundle exec htmlproofer --check-external-hash --check-html .
```

> Running ["ScriptCheck", "LinkCheck", "ImageCheck", "HtmlCheck"] on ["."] on *.html... 
> 
> 
> Checking 0 external links...
> Ran on 1 file!
> 
> 
> - ./index.html
>   *  linking to index.html#, but  does not exist (line 1)
>      \<a href="index.html#"\>link\</a\>
> htmlproofer 3.6.0 | Error:  HTML-Proofer found 1 failure!

ilyalyo:
It produce error with index.html exectly like this:
```html
<a href="https://travel.state.gov/content/dam/passports/FIG_cropper.swf#">
```
I thought its more general, but it works fine for all other links..

gjtorikian:
@ilyalyo Do you have the file with that link publicly available somewhere?

`string contains null byte` usually means that there's an encoding problem somewhere. HTML-Proofer expects UTF-8 encoded data, so probably for this one file, something is out of place. It's invisible to the naked eye. If I could get the specific file I could provide more information and probably have Proofer coerce input into UTF-8.

ilyalyo:
@gjtorikian file encoding is ascii.
You can download file with this link: https://drive.google.com/open?id=0B3_PHfM1xGNAbml2WGRXc3F6T2s

gjtorikian:
Can you convert the file to UTF-8?

ilyalyo:
@gjtorikian added PR

gjtorikian:
fixed by https://github.com/gjtorikian/html-proofer/pull/454

PR #454, opened 2017-09-30
Replace null chars with empty ones

ilyalyo:
issue #409

### DIFF
--- a/lib/html-proofer/url_validator.rb
+++ b/lib/html-proofer/url_validator.rb
@@ -175,7 +175,7 @@ module HTMLProofer
       return false unless @options[:check_external_hash]
       return false unless (hash = hash?(href))
 
-      body_doc = create_nokogiri(response.body)
+      body_doc = create_nokogiri(response.body.gsub("\u0000", ''))
 
       # user-content is a special addition by GitHub.
       xpath = %(//*[@name="#{hash}"]|//*[@id="#{hash}"])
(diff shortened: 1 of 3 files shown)


gjtorikian:
Could you add a test for this as well, please?

ilyalyo:
updated

fulldecent:
Looks good, I am +1 to merge

gjtorikian:
The test doesn't appear to be doing anything. I checked this branch out, and then reverted the one-line change made:

``` diff
diff --git a/lib/html-proofer/url_validator.rb b/lib/html-proofer/url_validator.rb
index 672e783..c54d010 100644
--- a/lib/html-proofer/url_validator.rb
+++ b/lib/html-proofer/url_validator.rb
@@ -175,7 +175,7 @@ module HTMLProofer
       return false unless @options[:check_external_hash]
       return false unless (hash = hash?(href))
 
-      body_doc = create_nokogiri(response.body.gsub("\u0000", ''))
+      body_doc = create_nokogiri(response.body)
 
       # user-content is a special addition by GitHub.
       xpath = %(//*[@name="#{hash}"]|//*[@id="#{hash}"])
```

The test provided still passes, which means that the line change as no effect on the test whatsoever.

That's because the file provided doesn't actually have a `null` byte in it. I had to go back and fetch [this file](https://github.com/gjtorikian/html-proofer/issues/409#issuecomment-303633639) before the test demonstrated a pass/fail.

Issue #447, opened 2017-08-26
Possible regression: htmlParseEntityRef: expecting ';' for protocol relative URLs


bascht:
First & foremost: Thank you so much for developing html-proofer. It's been a life saver for countless times. :heart_eyes_cat: 

I might have found a regression of #267 or #386:

### The issue

This snippet:

```html
<script src="//maps.googleapis.com/maps/api/js?callback=boing&key=xyz"></script>
```
will yield an error:

```log
*  822:87: ERROR: htmlParseEntityRef: expecting ';' (line x)
```
whereas this snippet will pass:
```html
<script src="https://maps.googleapis.com/maps/api/js?callback=boing&key=xyz"></script>
```

Let me know if I can supply more information or help to debug this issue. :smile:

gjtorikian:
Hmm, which version of `html-proofer` are you using? I just [pushed up a test](https://github.com/gjtorikian/html-proofer/commit/334a178ab456f43fa17bd6039888af538d8422c8), but [it passes](https://travis-ci.org/gjtorikian/html-proofer/jobs/268983770).

bascht:
Well that's interesting. I'm on 3.7.2. – a minimal example will pass, but if I wrap it into some HTML, it will fail:

```sh
# htmlproofer --version
htmlproofer 3.7.2
```

```html
# cat fail.html
<!doctype html>
<html>
<head>
<title>Test</title>
<script src="//maps.googleapis.com/maps/api/js?v=3.exp&callback=xyz&key=abc"></script>
</head>
</html>
# cat pass.html
<!doctype html>
<html>
<head>
<title>Test</title>
<script src="https://maps.googleapis.com/maps/api/js?v=3.exp&callback=xyz&key=abc"></script>
</head>
</html>
```
```sh
# htmlproofer --check-html fail.html
Running ["HtmlCheck", "LinkCheck", "ScriptCheck", "ImageCheck"] on fail.html on *.html...


Checking 1 external link...
Ran on 1 file!


- fail.html
  *  5:64: ERROR: htmlParseEntityRef: expecting ';' (line 5)
  *  5:72: ERROR: htmlParseEntityRef: expecting ';' (line 5)
htmlproofer 3.7.2 | Error:  HTML-Proofer found 2 failures!
```
```sh
# htmlproofer --check-html pass.html
Running ["HtmlCheck", "LinkCheck", "ScriptCheck", "ImageCheck"] on pass.html on *.html...


Checking 1 external link...
[...]

PR #461, opened 2017-11-28
Adjust regex to account for protocol-less links

gjtorikian:
Closes https://github.com/gjtorikian/html-proofer/issues/447.

Thanks for the report, @bascht!


bascht:
🎉 Awesome, thanks for the fix!

PR #398, opened 2017-03-06
Remove chomp URL from debug output

mattclegg:
It doesn't seem to make much sense to `chomp` the URL in the debug output.

If I'm looking at the debug logs and it's saying that a particular URL has an error. I expect it to tell me the exact string of the URL that has the error -not a string that has been manipulated.

PR #462, opened 2017-12-07
dropping url.chomp('/') which cause 301s

ldemailly:
And upgrading typhoeus

to solve https://github.com/gjtorikian/html-proofer/pull/398#issuecomment-349415894


Floppy:
So do we think that in the Typheous version updated here, the extraneous trailing / has been fixed? Or should we still be using `Addressable::URI#normalize`, rather than just removing the existing normalization code?

ldemailly:
I left unescape_url because I wasn't sure of what it was doing exactly

from running without / chomping in CI I see no issue - or rather - it fixes a different bug where missing / causes a 301 timeout (some other bug in curl version in circleci or typhoeus) so for me it's been a improvement without which my CI build never passes

PR #464, opened 2017-12-18
Update nokogiri to ~> 1.8.1 (CVE-2017-9050)

theneva:
Nokogiri 1.8.0 and below contains a vulnerability (http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-9050, https://security.gentoo.org/glsa/201711-01) reported by GitHub via e-mail today. This upgrades nokogiri to the new, patched version.

I'm not a Ruby expert, so this is mostly just to get the ball rolling. I don't think I would know what to do if it fails…


gjtorikian:
Thank you!

PR #469, opened 2018-03-20
Improve / Fix Example Code in README

nkuehn:
While trying out I found that the example code for a custom check in the README does not actually run.   It's mainly a word order typo.   

I added the check for nil to the oktopost example, too because it's suspectible to the same issues - but that's optional and could be left out for conciseness sake. 

If you want you can take over the .gitignore addition for JetBrains RubyMine IDE users, too. But that's just cosmetics.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ docs/
 _yardoc
 doc/
 .DS_Store
+.idea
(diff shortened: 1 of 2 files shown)


gjtorikian:
Great, thank you!

PR #478, opened 2018-06-02
Remove trailing whitespaces

LuizHAssuncao:
The build is broking on Rubocop because of this. So I'm removing these trailing whitespaces.


gjtorikian:
Gross! Thank you.

PR #479, opened 2018-07-09
Changed description of allow_hash_href option

iBobik:
Previous description is not clear about if it allows only "#" or also strings starting with "#". My version seems more clear for me. :-)


gjtorikian:
Could you please fix the description in `bin/htmlproof` as well? Thanks!

gjtorikian:
Thank you!

Issue #497, opened 2018-11-27
Links being changed causing hash to not be found


robertlove:
I have the following link:

```
<a href="https://docs.q-ctrl.com/control-operations#mølmer-sørensen-drive">Mølmer-Sørensen unitary</a>
```

And the following heading on the `/control-operations` page:

```
<h2 id="mølmer-sørensen-drive">Mølmer-Sørensen Drive</h2>
```

The link works. However, HTML Proofer gives the following error:

```
- ./_site/control-library.html
  *  External link https://docs.q-ctrl.com/control-operations#m%C3%B8lmer-s%C3%B8rensen-drive failed: https://docs.q-ctrl.com/control-operations exists, but the hash 'm%C3%B8lmer-s%C3%B8rensen-drive' does not
```

Note that `ø` is being changed to `%C3%B8` - which means it's now looking for `#m%C3%B8lmer-s%C3%B8rensen-drive` instead of `#mølmer-sørensen-drive`.

Issue #493, opened 2018-10-09
Checking anchors isn't working for me


plaindocs:
Apologies for the quasi support nature of this issue.

I'm having trouble checking (internal) anchors in the Travis CI docs, even with `check_external_hash: true`. 

Most of the internal links are relative, due to netlify preview urls, so not sure that the internal domain setting is doing anything. 

A lot of them are case changes only (such as https://github.com/travis-ci/docs-travis-ci-com/pull/2050/files ) 

If you have any ideas on what I'm doing wrong that would be awesome. 

```
desc 'Runs the html-proofer test'
task :run_html_proofer => [:build] do
  HTMLProofer.check_directory(
    './_site',
    internal_domains: ['docs.travis-ci.com'],
    check_external_hash: true,
    check_html: true,
    connecttimeout: 600,
    allow_hash_ref: true,
    only_4xx: true,
    typhoeus: {
      ssl_verifypeer: false, ssl_verifyhost: 0, followlocation: true
    },
  ).run
end
```

gjtorikian:
https://github.com/travis-ci/docs-travis-ci-com/pull/2050 is a passing build. Could you show me a failing build where this is occurring?

plaindocs:
Hey @gjtorikian 

Thanks for the ping back.

I'm made a PR reverting that fix:

https://github.com/travis-ci/docs-travis-ci-com/pull/2052

The link from the home page is (which does not exist):

https://5bbe13adfdd72a52fbcf4c30--musing-lamport-7fb66c.netlify.com/user/notifications/#Changing-the-email-address-for-build-notifications 

And the actual anchor (which does exist) is: 

https://5bbe13adfdd72a52fbcf4c30--musing-lamport-7fb66c.netlify.com/user/notifications/#changing-the-email-address-for-build-notifications

htmlproofer does not report any errors: 

https://travis-ci.org/travis-ci/docs-travis-ci-com/builds/439683402#L646

PR #498, opened 2018-11-28
Check hash capitals

gjtorikian:
Previously, the case-sensitivity of hashes was not being considered. I believe that this is important for browsers--and if not, it's important for correctness!. This PR now asserts that the case sensitivity of a hash _is_ important.

As well, it turns out that some of the tests weren't running at all, due to missing files! Previously, proofer didn't check if a file/directory existed before any validation. Ugh. This PR also fixes that, along with a boatload of broken tests.

Closes https://github.com/gjtorikian/html-proofer/issues/493, closes https://github.com/gjtorikian/html-proofer/issues/497


plaindocs:
You star! :star2: :star2: :star2: :star2: Thank you!

Of course, this does mean that I get to fix 133 broken links now :trollface:

gjtorikian:
Yeah I thought about making it an option (`--enforce-case-sensitivity`?) but I'd rather opt for strict correctness here.

PR #499, opened 2018-12-12
Ignoring sri for pagination rels 

### LINKED ISSUE #495: SRI and CORS errors in <link> tag where they shouldn't be
Hello.

Recently I've run into a problem when using seo and paginator plugins. The seo plugin outputs a ```<link rel="next">``` or ```<link rel="prev>``` tag with a proper link to another page. 

The problem is, html-proofer is pointing out the "SRI and CORS not provided" error if the link is different from the page it is on. As far as I've looked, I can't change the links from the absolute form to prevent this. I believe this error shouldn't even pop up in this case. What should I do to fix this?

vllur:
Will fix #495.
Kinda blowed away my previous fork (https://github.com/gjtorikian/html-proofer/pull/496#issue-228174884) - I wanted to write some checks around pagination links, but I don't have setup to run tests locally (maybe when I'll have normal setup I'll try to do this). 

I've added pagination rels to ignore and writed simple test for it. @Floppy, what do you think?

### DIFF
--- a/lib/html-proofer/check/links.rb
+++ b/lib/html-proofer/check/links.rb
@@ -127,7 +127,7 @@ class LinkCheck < ::HTMLProofer::Check
     html.xpath(*xpaths)
   end
 
-  IGNORABE_REL = %(canonical alternate icon manifest apple-touch-icon)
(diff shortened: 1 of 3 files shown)


gjtorikian:
Looks good to me, thank you.

PR #503, opened 2018-12-29
Make assume_extension work with a directory with the same name

### LINKED ISSUE #502: assume_extension fails when a directory with the same name as a file exists
If a link refers to a file without an extension, and there exists a directory with the same name as that file (without the extension), html-proofer will complain about that link. For example, if I run `htmlproofer --assume-extension` on:

```html
<!-- site/page.html -->
<a href="test">test</a>
```

and a file named `test.html` exists *and* a directory named `test` both exist, html-proofer will claim that test does not exist.

nex3:
Closes #502

### DIFF
--- a/lib/html-proofer/element.rb
+++ b/lib/html-proofer/element.rb
@@ -173,11 +173,10 @@ module HTMLProofer
 
       file = File.join base, path
 
-      # implicit index support
-      if File.directory?(file) && !unslashed_directory?(file)
-        file = File.join file, @check.options[:directory_index_file]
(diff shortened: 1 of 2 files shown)


gjtorikian:
Hi, this looks good, thank you for the change!

I see the fixtures files were added (`same_name_as_dir.html`), but could you also add a corresponding test to ensure this doesn't break in the future? Thanks!

nex3:
There's an existing test that no errors are raised for `no_html_extension.html`. I verified that the test failed before I added my fix.

gjtorikian:
Oh derp, of course, it's just running through all the links in that file.

Thanks again!

Issue #506, opened 2019-01-19
False positives on img with new lines on src value


cetinajero:
Hi Garen,

I came across an issue when an img tag has this structure:

```html
<img src="
https://grupopv-public.s3.amazonaws.com/assets/img/catalog/thumbnails/radios/motorola/licenses/smart-ptt/spttc0003.png
" class="img-center" alt="SPTTC0003" />

```

As you can see there are new lines at the _start_ and _end_ of the _src value_, and if you see closely there are even some spaces too, this is a common scenario when using [Liquid](https://shopify.github.io/liquid/) variables, tags & filters.

```bash
Edgars-MacBook-Pro:spttc0003 cetinajero$ cat index.html | od -c
0000000                                            <   i   m   g       s
0000020    r   c   =   "  \n                                           h
0000040    t   t   p   s   :   /   /   g   r   u   p   o   p   v   -   p
0000060    u   b   l   i   c   .   s   3   .   a   m   a   z   o   n   a
0000100    w   s   .   c   o   m   /   a   s   s   e   t   s   /   i   m
0000120    g   /   c   a   t   a   l   o   g   /   t   h   u   m   b   n
0000140    a   i   l   s   /   r   a   d   i   o   s   /   m   o   t   o
0000160    r   o   l   a   /   l   i   c   e   n   s   e   s   /   s   m
0000200    a   r   t   -   p   t   t   /   s   p   t   t   c   0   0   0
0000220    3   .   p   n   g  \n                                   "    
0000240    c   l   a   s   s   =   "   i   m   g   -   c   e   n   t   e
0000260    r   "       a   l   t   =   "   S   P   T   T   C   0   0   0
0000300    3   "       /   >  \n                                        
0000306
```

When html-proofer test a node like this it fails in reporting errors and because of this, my CI tests are reporting false positives.

gjtorikian:
Closed via https://github.com/gjtorikian/html-proofer/commit/c8d8b9bd5ca63d7bcc20e2c29f2e18211e152f6b#diff-42adaf791dc0552cebbfe8060b2fe714R57

PR #507, opened 2019-01-19
Strip new lines and spaces from img urls

cetinajero:
Fixes https://github.com/gjtorikian/html-proofer/issues/506

### DIFF
--- a/lib/html-proofer/check/images.rb
+++ b/lib/html-proofer/check/images.rb
@@ -13,6 +13,10 @@ class ImageCheck < ::HTMLProofer::Check
     blank?(@img.url)
   end
 
+  def clean_url
+    @img.url.strip!
+  end
+
   def run
     @html.css('img').each do |node|
       @img = create_element(node)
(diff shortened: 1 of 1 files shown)


gjtorikian:
Hi! Thanks for the PR. Normally I would accept these, but, there is actually a spot "higher up" in the code where the `strip` can be inserted: https://github.com/gjtorikian/html-proofer/commit/c8d8b9bd5ca63d7bcc20e2c29f2e18211e152f6b#diff-42adaf791dc0552cebbfe8060b2fe714R57

html-proofer keeps track of `src` for `img` and `href` for `a` tags and just calls them `url`s. I imagine this same problem will affect HTML anchors, so it makes sense to me to put it there.

I've released that fix as 3.10.2. Thanks for the report!

cetinajero:
> html-proofer keeps track of `src` for `img` and `href` for `a` tags and just calls them `url`s. I imagine this same problem will affect HTML anchors, so it makes sense to me to put it there.

Totally agree 👍 👍 

> I've released that fix as 3.10.2. Thanks for the report!

Thanks to you too!

PR #512, opened 2019-03-14
Add rack middleware for proofing HTML at runtime

DanielHeath:
Sample usage in a Rails app: Add to `config/application.rb`

    config.middleware.use HTMLProofer::Middleware if Rails.env.test?
    config.middleware.use HTMLProofer::Middleware if Rails.env.development?

This will raise an error at runtime if your HTML is invalid. 

Particularly helpful for projects which have extensive CI.

### DIFF
--- a/lib/html-proofer/middleware.rb
+++ b/lib/html-proofer/middleware.rb
@@ -1,0 +1,65 @@
+
+
+module HTMLProofer
+  class Middleware
+
+    def self.options
+      @options ||= {
+        type:                :file,
+        allow_missing_href:  true, # Permitted in html5
+        allow_hash_href:     true,
+        check_external_hash: true,
+        check_html:          true,
+        url_ignore:          [/.*/], # Don't try to check local files exist
+      }
+    end
+
+    def initialize app
+      @app = app
+    end
+
+    HTML_SIGNATURE = [
+      "<!DOCTYPE HTML",
+      "<HTML",
+      "<HEAD",
+      "<SCRIPT",
+      "<IFRAME",
+      "<H1",
+      "<DIV",
+      "<FONT",
+      "<TABLE",
+      "<A",
+      "<STYLE",
+      "<TITLE",
+      "<B",
+      "<BODY",
+      "<BR",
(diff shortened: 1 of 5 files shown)


lib/html-proofer/middleware.rb
```diff
@@ -0,0 +1,66 @@
...
+      '<TABLE',
+      '<A',
+      '<STYLE',
+      '<TITLE',
+      '<B',
+      '<BODY',
+      '<BR',
+      '<P',
+      '<!--'
+    ]
```
gjtorikian: What is this subset based on? Is there a more definite way to get a list (eg., does Nokogiri provide a constant of tag names? What happens if a page has an `<h2` instead of `<h1`?)

lib/html-proofer/middleware.rb
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+        if parsed[:failures].length > 0
+          raise "HTML Validation errors (skip by adding ?SKIP_VALIDATION to URL): \n#{parsed[:failures].join("\n")}"
+        end
+      end
+      result
+    end
+  end
+end
```
gjtorikian: Adding a test and documentation in the README for this middleware would be helpful!

lib/html-proofer/middleware.rb
DanielHeath: This is drawn from the golang mime-type sniffing package. 

Almost every page is going to start with `<!DOCTYPE HTML` or a comment.

DanielHeath: Note that it's checking for 'starts with', not 'contains'. 

If your entire document starts with an `<h2` it's not going to fire, but I think at that point you already know your html is invalid.

lib/html-proofer/middleware.rb
DanielHeath: Done!

DanielHeath:
@gjtorikian I think that was everything - anything else needed before a merge?

gjtorikian:
Thank you!

PR #523, opened 2019-07-04
Making runner return the array of failures to enable file export

altcointrading:
If runner returns the data like this, in scripts some variation of the following can be done:

```
def htmlproof path
  options = { :assume_extension => true }
  errorfile = "#{path}/errors.csv"
  if File.directory?(path)
      res = HTMLProofer.check_directory(path, options).run
      `echo '' > #{errorfile}`
      CSV.open(errorfile, "a") do |csv|
        res.each do |row|
          csv << [row]
        end
      end
  end
end
```

The result is a CSV with one failure per row.

It would be really really handy to have the option to export the failures to a CSV!!

### DIFF
--- a/lib/html-proofer/runner.rb
+++ b/lib/html-proofer/runner.rb
@@ -167,9 +167,7 @@ module HTMLProofer
       sorted_failures = SortedIssues.new(@failures, @options[:error_sort], @logger)
 
       sorted_failures.sort_and_report
-      count = @failures.length
-      failure_text = pluralize(count, 'failure', 'failures')
-      raise @logger.colorize :red, "HTML-Proofer found #{failure_text}!"
+      return @failures
     end
   end
 end


lib/html-proofer/runner.rb
```diff
@@ -167,9 +167,7 @@ def print_failed_tests
       sorted_failures = SortedIssues.new(@failures, @options[:error_sort], @logger)
 
       sorted_failures.sort_and_report
-      count = @failures.length
-      failure_text = pluralize(count, 'failure', 'failures')
-      raise @logger.colorize :red, "HTML-Proofer found #{failure_text}!"
+      return @failures
```
gjtorikian: Sorry, but you are removing the existing use case of printing errors on the command line for the sake of your specific use case of iterating over the failures in a script. 

An option should be passed to enable the returning of these failures, or at least some TTY detection to note that when this is run from the terminal, the logging is the option.

altcointrading:
Thanks for the feedback. 

I will add a tty detection routine and a prompt to choose whether to write the result to a file in selected location.

gjtorikian:
I think this was superseded by https://github.com/gjtorikian/html-proofer/pull/538

PR #528, opened 2019-08-12
add whitelisted rel types for indieweb links

Floppy:
Various [indieweb](https://indieweb.org/) capabilities use `link` tags, with `rel`s like `me`, `webmention` and `pingback`. Those don't need SRI or CORS, so this PR whitelists them.

I think there's a better approach to this using a blacklist of rel types rather than a whitelist - after all, the browser will only load certain types. I'll put together another PR for that (update, it's https://github.com/gjtorikian/html-proofer/pull/529), but I wanted to get this simple fix in first in case that one proves more contentious.

If #529 is the preferred solution, this can either be binned or I can update it to just include the new test.

### DIFF
--- a/spec/html-proofer/fixtures/links/link_with_me.html
+++ b/spec/html-proofer/fixtures/links/link_with_me.html
@@ -1,0 +1,9 @@
+<html>
+<head>
+    <link rel="me" href="https://github.com/gjtorikian/html-proofer"/>
+    <link rel="webmention" href="https://webmention.io/username/webmention" />
+    <link rel="pingback" href="https://webmention.io/username/xmlrpc" />
+</head>
+<body>
+</body>
+</html>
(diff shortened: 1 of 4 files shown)


spec/html-proofer/fixtures/links/link_with_me.html
```diff
@@ -0,0 +1,9 @@
+<html>
+<head>
+    <link rel="me" href="https://github.com/gjtorikian/html-proofer"/>
+    <link rel="webmention" href="https://webmention.io/username/webmention" />
+    <link rel="pingback" href="https://webmention.io/username/xmlrpc" />
```
gjtorikian: Haven't decided yet if this or #529 are preferable, but these should all be faked/broken/garbage links, to prove that they are being skipped.

Floppy: good point, I can fix that if you prefer this solution over #529.

fulldecent:
Preferring #529 over this since it better matches the specification, which is a whitelist.

gjtorikian:
I trust @fulldecent.

PR #529, opened 2019-08-12
Check only stylesheet link rels rather than whitelisting other rels

Floppy:
This is the alternative solution for the addition in #528. The SRI spec only applies to `<link rel='stylesheet'>`, so this PR only applies the check to those. This means we don't need an ever-expanding whitelist of rel types, but can easily add more into the check.

### DIFF
--- a/lib/html-proofer/check/links.rb
+++ b/lib/html-proofer/check/links.rb
@@ -129,6 +129,8 @@ class LinkCheck < ::HTMLProofer::Check
     html.xpath(*xpaths)
   end
 
+  # Whitelist for affected elements from Subresource Integrity specification
+  # https://w3c.github.io/webappsec-subresource-integrity/#link-element-for-stylesheets
   SRI_REL_TYPES = %(stylesheet)
 
   def check_sri(line, content)
--- a/lib/html-proofer/check/links.rb
+++ b/lib/html-proofer/check/links.rb
@@ -129,10 +129,10 @@ class LinkCheck < ::HTMLProofer::Check
     html.xpath(*xpaths)
   end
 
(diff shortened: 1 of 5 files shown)


lib/html-proofer/check/links.rb
```diff
@@ -129,10 +129,10 @@ def find_fragments(html, fragment_ids)
     html.xpath(*xpaths)
   end
 
-  IGNORABE_REL = %(canonical alternate next prev previous icon manifest apple-touch-icon)
+  SRI_REL_TYPES = %(stylesheet)
 
   def check_sri(line, content)
-    return if IGNORABE_REL.include?(@link.rel)
+    return unless SRI_REL_TYPES.include?(@link.rel)
```
gjtorikian: I like this but I think I have to think on it a bit more before approving. For example, `script` and `link` will also need to be checked. I'm not sure if there are more.

Floppy: So, this code should only apply to `link` tags - `script` is handled by another check, in https://github.com/gjtorikian/html-proofer/blob/master/lib/html-proofer/check/scripts.rb#L33, so is unaffected. As I mentioned, as far as I can tell from the spec, the only thing that is proactively loaded in by browsers from `link` tags are `rel='stylesheet'`, which this should cover.

fulldecent:
@gjtorikian Does this additional documentation resolve the issue? 277e1c934313f45df086a9fe0d3a98ede95d8808

lib/html-proofer/check/links.rb
gjtorikian: You're right! My mistake.

gjtorikian:
Thanks y'all!

PR #530, opened 2019-08-20
Update "activesupport" dependency to allow Rails 6

cpruitt:
With Rails 6.0 released we're encountering some dependency conflicts with html-proofer being tied to active support `< 6.0`.

This PR bumps the upper bound of the ActiveSupport dependency version in `html-proofer.gemspec` from `< 6.0` to `< 7.0`.

### DIFF
--- a/html-proofer.gemspec
+++ b/html-proofer.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency 'yell',            '~> 2.0'
   gem.add_dependency 'parallel',        '~> 1.3'
   gem.add_dependency 'addressable',     '~> 2.3'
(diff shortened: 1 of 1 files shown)


Floppy:
Tests ran fine with activesupport 6.0, so I think this will be fine - my only concern is that rails 6 requires Ruby 2.5.0 or higher, so we might want to specify the same thing.

Floppy:
Actually, we don't specify a ruby requirement anywhere, other than the list that we test against in `.travis.yml`, so as long as those pass, we should be fine.

gjtorikian:
Hi, thanks for this! This AR bump happens every time a new Rails release is out, so I opened up https://github.com/gjtorikian/html-proofer/pull/534 to instead just remove the very limited usage of ActiveRecord. Thanks for your contribution, though, I'll bring a new release out shortly.

PR #531, opened 2019-08-24
DO NOT MERGE: uppercase paths failing?

afeld:
Running against a static site, where HTMLProofer was reporting broken links if the links/files with uppercase characters. Weirdly, this was only happening in some places (CI, on Linux) and not others (my machine, a Mac). I then assumed it must be an issue with case-sensitive vs. insensitive filesystems, but since this test I added passes in (case-sensitive) Linux in TravisCI, I am totally confused about what's going on. Ideas?

### DIFF
--- a/spec/html-proofer/fixtures/links/link_mismatched_case.html
+++ b/spec/html-proofer/fixtures/links/link_mismatched_case.html
@@ -1,0 +1,1 @@
+<a href="ancHors_iN_pRe.html">Relative to self</a>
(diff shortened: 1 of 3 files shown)


gjtorikian:
Hi @afeld! 

I only followed [one link](https://handbook.18f.gov/travel-101/), and the section stating "[Rebooking and changing official travel plans](https://handbook.18f.gov/travel-guide-A-amended-authorizations/)" does indeed have a broken link. It seems that the prefix `policies/travel-and-leave-policies` is not being included?

I think this is a problem on your build, so I will close this issue, but please let me know if I am wrong and reopen this!

afeld:
Sorry, you're right - something weird is happening, but it's on the site/build side. Thanks!

PR #538, opened 2019-09-25
Allow accessing list of issues directly

muan:
👋 This pull request adds a reader for `@failures` so users can access them directly, and use the `Issue` to construct their report if they wish to do so, like what #523 wanted to do. 

In our use case, we mostly want to avoid Proofer from raising a generic `RuntimeError`, and so we will be printing out the errors ourselves. I initially accessed `#failed_tests` for this, but found that we'd want to group results with path. Since `#failed_test` gives us `issue.to_s`, we will have to manually parse it back. Having `#failures` would be more convenient and gives developers flexibility around reporting.

It'd also be good to decouple the checks in `#run` into another method so it can be called without raise/logger. I thought I'd start small for now. Thoughts?

cc @gjtorikian 😬

### DIFF
--- a/lib/html-proofer/runner.rb
+++ b/lib/html-proofer/runner.rb
@@ -4,7 +4,7 @@ module HTMLProofer
   class Runner
     include HTMLProofer::Utils
 
-    attr_reader :options, :external_urls
+    attr_reader :options, :external_urls, :failures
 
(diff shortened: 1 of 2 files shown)


gjtorikian:
Sure, this works for me. I'll release this as a new minor version of the library soon. Thanks!

PR #539, opened 2019-10-01
Fix: internal link misrepresented and misused

Graborg:
I found some confusing code and followed it down the rabbit hole. I hope this will make the code more maintainable.

### DIFF
--- a/lib/html-proofer/check/links.rb
+++ b/lib/html-proofer/check/links.rb
@@ -79,7 +79,7 @@ class LinkCheck < ::HTMLProofer::Check
   end
 
   def handle_mailto(link, line, content)
-    if link.path.nil?
+    if link.path.empty?
       add_issue("#{link.href} contains no email address", line: line, content: content)
     elsif !link.path.include?('@')
       add_issue("#{link.href} contains an invalid email address", line: line, content: content)
@@ -87,7 +87,7 @@ class LinkCheck < ::HTMLProofer::Check
   end
 
   def handle_tel(link, line, content)
-    add_issue("#{link.href} contains no phone number", line: line, content: content) if link.path.nil?
+    add_issue("#{link.href} contains no phone number", line: line, content: content) if link.path.empty?
   end
 
   def handle_hash(link, line, content)
--- a/lib/html-proofer/check/links.rb
+++ b/lib/html-proofer/check/links.rb
@@ -49,7 +49,7 @@ class LinkCheck < ::HTMLProofer::Check
         next if @link.respond_to?(:rel) && @link.rel == 'dns-prefetch'
         add_to_external_urls(@link.href)
         next
-      elsif !@link.internal? && !@link.exists?
+      elsif !@link.remote? && !@link.exists?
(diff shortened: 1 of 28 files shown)


lib/html-proofer/check/links.rb
```diff
@@ -49,7 +49,7 @@ def run
         next if @link.respond_to?(:rel) && @link.rel == 'dns-prefetch'
         add_to_external_urls(@link.href)
         next
-      elsif !@link.internal? && !@link.exists?
+      elsif !@link.remote? && !@link.exists?
```
gjtorikian: Wouldn't this be best as `@link.internal?` (no `!`)?

lib/html-proofer/check/links.rb
```diff
@@ -79,15 +79,15 @@ def check_schemes(link, line, content)
   end
 
   def handle_mailto(link, line, content)
-    if link.path.empty?
+    if link.path.nil?
```
gjtorikian: For this and L90, I'd prefer the revert. I believe In the even of the path being missing, the `path` attribute is kept as `""`, which, while not `nil`, is an empty string, and that is wrong in my opinion.

lib/html-proofer/element.rb
```diff
@@ -151,11 +151,11 @@ def param_link
     end
 
     def slash_link
-      url.start_with?('|')
+      url.start_with?('/')
```
gjtorikian: ....what the hell was I thinking here? o_O

lib/html-proofer/element.rb
```diff
@@ -221,9 +220,9 @@ def base
 
     def html
       # If link is on the same page, then URL is on the current page so can use the same HTML as for current page
-      if (hash_link || param_link) && internal?
+      if (hash_link || param_link)
         @html
-      elsif slash_link && internal?
+      elsif slash_link
```
gjtorikian: Nice cleanup, thanks!

lib/html-proofer/check/links.rb
Graborg: It would. I tried it and it created some more errors failing. Looks to be some discrepancy there. I will give another go.

Graborg:
Tuns out https://github.com/gjtorikian/html-proofer/pull/539#discussion_r330193842 created quite the domino effect, it deserves another PR, if thats okay with you.

gjtorikian:
Thanks!

lib/html-proofer/element.rb
```diff
@@ -77,11 +79,11 @@ def path
     end
 
     def hash
-      parts.fragment unless parts.nil?
+      parts&.fragment
```
shushugah: Wouldn't safe navigation fail for ruby < 2.3.0? and if so, is a minimum version of ruby specified anywhere? cc:	@gjtorikian

gjtorikian: < 2.3 is EOL (in fact, 2.4 is EOL soon) and I don't support dead rubies.

PR #541, opened 2019-10-03
refactor: use internal instead of not external

Graborg:
Based on #539 

Change from !remote? -> internal? and handle all the domino effects of this.


gjtorikian:
Thanks!

PR #577, opened 2020-08-31
Add HTMLProofer::Runner#before_request

### LINKED ISSUE #86: Support links behind auth
Occasionally, at GitHub, we'll link to sites within github.com that are behind auth. For example: `[check out this discussion](https://www.github.com/github/secret-internal-repo/issues/23)`.

We've had to exclude these links by writing them out as HTML and adding `data-proofer-ignore`. Blah.

I think instead there should be a new config option hash that takes a domain as a key, and an OAuth token as the value, so that these sorts of links can be checked. For example, you'd pass in `:domain_auth => { "github.com" => ENV['MACHINE_USER_TOKEN'] }`. When HTML::Proofer hits a 404, it'd look the domain up, and try to use the provided token to recheck the link.  

/cc @penibelst @parkr Y'all think this makes sense?

asbjornu:
This PR adds the `HTMLProofer::Runner#before_request` method so the `Typhoeus::Request` object can be modified before the request is executed. I'm not happy about the accompanying spec, so any input on how to improve it is welcome.

Fixes #86.

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,20 @@ The default value is:
 }
 ```
 
+#### Setting `before-request` callback
+
+You can provide a block to set some logic before an external link is checked. For example, say you want to provide an authentication token every time a GitHub URL is checked. You can do that like this:
+
+```ruby
+proofer = HTMLProofer.check_directory(item, opts)
+proofer.before_request do |request|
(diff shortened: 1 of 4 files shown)


asbjornu:
Ping @gjtorikian

gjtorikian:
Thank you for this! It'll go out now as v 3.16.

Issue #576, opened 2020-08-30
data-proofer-ignore not helping to filter out error about `@click` anchor attribute


zorn:
I love using the command line `htmlproofer` to help check my site but today as I added an Alpine.js behavior I got this error:

```
#!/bin/bash -eo pipefail
htmlproofer ~/project/public --allow-hash-href --check-html --empty-alt-ignore --disable-external

Running ["ScriptCheck", "LinkCheck", "ImageCheck", "HtmlCheck"] on ["/home/circleci/project/public"] on *.html... 


Attribute set `<a href="#" @click="showIntroVideo = !showIntroVideo" onclick="stopVideo(); return false;" data-proofer-ignore="true">
                            Close
                        </a>` contains an error!
htmlproofer 3.15.3 | Error:  invalid attribute name `@click'

Exited with code exit status 1
CircleCI received exit code 1
```

I'm trying to use `data-proofer-ignore` to filter out the issue but it doesn't seem to work. Any suggestions?

gjtorikian:
Sorry about that! This should be fixed in the next patch release. Thanks for the report!

zorn:
Thanks so much for the help!!

PR #578, opened 2020-09-11
Allow for at-signs

gjtorikian:
Closes https://github.com/gjtorikian/html-proofer/issues/576


gjtorikian:
Failure is from Rubocop, no time to fix at the moment.

PR #590, opened 2020-11-14
support cache for internal links

### LINKED ISSUE #579: HTMLProofer runs out of memory
Hello, we are using HTMLProofer to verify our [built web site](https://github.com/SwedbankPay/developer.swedbankpay.com), and it has been working great, however a week or so ago we started running into an issue where having too many pages makes Docker locally and on GitHub run out of memory. Not much trace, but a run with it can be seen [here](https://github.com/SwedbankPay/developer.swedbankpay.com/runs/1081305520?check_suite_focus=true#step:3:195).
We are not calling `HTMLProofer` directly, it's done via our Docker image -> [here](https://github.com/SwedbankPay/jekyll-plantuml-docker/blob/master/.docker/entrypoint/lib/commands/verifier.rb#L36) <-

If I delete 5-6 pages from the site it runs fine and is done in under a minute, but it seems like I've hit a solid wall where we have either too many pages or there are too many links to check.

I tested disabling the `LinkCheck` and it ran to completion in a few seconds.

Ruby is not my strongest suit, but I've tried to debug and reduce memory usage in a fork, but I've had no luck so far.
Running it in Parallel does not change the result, it only makes it crash/run out of memory faster.

Is there a setting I can change, or do you perhaps have a pointer to where I should continue looking to resolve this issue?

gjtorikian:
Should close #579


gjtorikian:
Test pass, failure is Rubocop-related, will fix soon

Issue #589, opened 2020-11-12
URLs in <source src=xx> not being checked.


argos83:
Hi!

URLs for embedded media are not being checked.

E.g.

```
<audio controls>
    <source src="THIS IS NOT CHECKED" type="audio/mpeg">
</audio>
```

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source

I'm using the latest version (at the moment of writing) 3.16.0

PR #591, opened 2020-11-14
test source tags

gjtorikian:
Should close https://github.com/gjtorikian/html-proofer/issues/589


gjtorikian:
Tests pass, Rubocop fails, will deal with that 🔜

PR #597, opened 2020-11-23
Fix internal link handling for dirs w/o root_dir

gjtorikian:


gjtorikian:
Travis seems to have stalled; about half the Ruby versions are green and the other half are queued. I'm considering this good to go.

PR #598, opened 2020-11-23
CI: Update Travis build matrix Ruby versions

olleolleolle:
I used versions mentioned in https://github.com/rvm/rvm/blob/master/config/known#L12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: ruby
 rvm:
-  - 2.4.3
-  - 2.5.0
-  - 2.6.0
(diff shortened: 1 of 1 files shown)


gjtorikian:
Thanks for this; I'm going to start using GitHub Actions because I'm fed up with how slow Travis is. I've migrated your suggestions here https://github.com/gjtorikian/html-proofer/commit/a77471424152764c7db4d920e0278611c0270e25

olleolleolle:
@gjtorikian I hope you'll get the speed gains!

PR #613, opened 2020-12-27
Allow usage of Ruby 3.0

tisba:
[Ruby 3.0 has been released](https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/) 🎉 

This PR makes the following changes:

* add Ruby 3.0 to build matrix
* use Ruby 3.0 for linting
* changes `gem.required_ruby_version` to `>= 2.4.10` and `< 4.0`.

~~The changes `gem.required_ruby_version` are TBD. Allowing everything above `2.4.10` (oldest Ruby in build matrix) looked like the simplest solution.~~

### DIFF
--- a/html-proofer.gemspec
+++ b/html-proofer.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.executables   = all_files.grep(%r{^bin/}) { |f| File.basename(f) }
   gem.test_files    = gem.files.grep(%r{^(spec)/})
   gem.require_paths = ['lib']
-  gem.required_ruby_version = '>= 2.4.10'
+  gem.required_ruby_version = ['>= 2.4.10', '< 4.0']
 
   gem.add_dependency 'addressable',     '~> 2.3'
   gem.add_dependency 'mercenary',       '~> 0.3'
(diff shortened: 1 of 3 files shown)


html-proofer.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.executables   = all_files.grep(%r{^bin/}) { |f| File.basename(f) }
   gem.test_files    = gem.files.grep(%r{^(spec)/})
   gem.require_paths = ['lib']
-  gem.required_ruby_version = '~> 2.4'
+  gem.required_ruby_version = '>= 2.4.10'
```
gjtorikian, line 20: I'm glad 3.0 passes but this isn't future-proof enough. Could you please change it to `'~> 2.4, <~ 3.0'` (or whatever the syntax for version ranges is, I always forget--basically >= 2.4 but < 4.0).

tisba: Okay, I'd go with

```suggestion
  gem.required_ruby_version = ['>= 2.4.10', '< 4.0']
```

btw: [Ruby 2.4 is EOL since 2020-04-01](https://www.ruby-lang.org/en/news/2019/04/01/ruby-2-4-6-released/), I'd recommend to drop support soonish.

gjtorikian:
Thanks!

html-proofer.gemspec
lizzby: 👍

PR #630, opened 2021-03-20
Proposal: Add an option to provide a data attribute as an image src

andrewbredow:
Hey there @gjtorikian! I'm using `html-proofer` in a project that uses `data-src` attributes on images to load optimized versions after page load. We're specifically using [UIKit](https://getuikit.com/docs/image#usage) to accomplish this, but it's also a common pattern in other libraries that manipulate image src on pageload.

In some cases there are both `src` attributes and `data-src` together. But we're also using a pattern that applies the default image to the background, so some image tags have `data-src` attributes but not `src` attributes when the initial HTML is rendered - which results in invalid image src errors. Rather than ignoring all of these images, my goal was to still have validation of the URLs, so I created a fork with a proof of concept for adding data attributes as valid image src urls via configuration.

I've really only tested the happy path here, and I'm not convinced the naming conventions are super clear, but thought I'd open this PR to see if you're interested in this feature making it's way into the core library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,7 @@ The `HTMLProofer` constructor takes an optional hash of additional options:
 | `check_opengraph` | Enables the Open Graph checker. | `false` |
 | `check_html` | Enables HTML validation errors from Nokogumbo | `false` |
 | `check_img_http` | Fails an image if it's marked as `http` | `false` |
+| `data_src_attribute` | Sets a data attribute to use as a valid image src, for example `data-src` | "" |
 | `check_sri` | Check that `<link>` and `<script>` external resources use SRI |false |
 | `checks_to_ignore`| An array of Strings indicating which checks you do not want to run | `[]`
 | `directory_index_file` | Sets the file to look for when a link refers to a directory. | `index.html` |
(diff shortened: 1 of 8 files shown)


gjtorikian:
This makes sense to me. I haven’t looked too carefully at the change just yet, but would it make sense to keep this as a Boolean flag that always just checks `data-sec`? In other words, how likely is it that some other data-* attribute would be the real source? (I know nothing about UIKit and whether this attribute is customizable.)

andrewbredow:
> In other words, how likely is it that some other data-* attribute would be the real source?

I think the only reason I erred toward a string option was that data attributes are the wild west 😂 I did a brief search and `data-src` seems to be a very common convention. It does seem like building it as a boolean option would be a great start, and certainly covers the common use cases. I'll update this PR with that change.

bkeepers:
I've seen `data-src` used on a couple different projects, and I don't recall seeing other attributes used for the same purpose.

Floppy:
I like the flexibility of making it customisable. Perhaps a middle ground is that if it was just set to `true`, it would use `data-src` by default?

gjtorikian:
Yes, I guess I prefer the customizable one too—and definitely that over an option that can be both a string and a Boolean. 

Maybe the this option should just be `img_src_attribute`, default to `src`, and if it’s set to something else, then that attribute is used instead.

Floppy:
I think in this use case though, `src` and `data-src` would both need to be checked, so setting it to check `data-src` shouldn't disable checking `src`.

andrewbredow:
@Floppy Yea, I think that's right... generally an image would have at least 1 src, but possibly more. Sheesh the web is complicated :)

aekewu (approved):

gjtorikian:
I'm not at all opposed to this PR, but I'm going to close it, because I've completely rewritten the internal logic (https://github.com/gjtorikian/html-proofer/pull/672) to prepare for the next major html-proofer release (https://github.com/gjtorikian/html-proofer/issues/669).

As far as I understand it would be desirable to keep this customizable, rather than locked into `data-src`. I'll keep that in mind while adding the feature. Sorry for the delay, thanks for your patience.

PR #632, opened 2021-03-29
Add --hydra-config flag to override concurrency from CLI

patcon:
Related: https://github.com/gjtorikian/html-proofer/issues/131

Thanks for considering, and thanks for all your work on maintaining this tool! :)

### DIFF
--- a/bin/htmlproofer
+++ b/bin/htmlproofer
@@ -48,6 +48,7 @@ Mercenary.program(:htmlproofer) do |p|
   p.option 'storage_dir', '--storage-dir PATH', String, 'Directory where to store the cache log (default: "tmp/.htmlproofer")'
   p.option 'timeframe', '--timeframe <time>', String, 'A string representing the caching timeframe.'
   p.option 'typhoeus_config', '--typhoeus-config CONFIG', String, 'JSON-formatted string of Typhoeus config. Will override the html-proofer defaults.'
+  p.option 'hydra_config', '--hydra-config CONFIG', String, 'JSON-formatted string of Hydra config. Will override the html-proofer defaults.'
   p.option 'url_ignore', '--url-ignore link1,[link2,...]', Array, 'A comma-separated list of Strings or RegExps containing URLs that are safe to ignore. It affects all HTML attributes. Note that non-HTTP(S) URIs are always ignored'
   p.option 'url_swap', '--url-swap re:string,[re:string,...]', Array, 'A comma-separated list containing key-value pairs of `RegExp => String`. It transforms URLs that match `RegExp` into `String` via `gsub`. The escape sequences `\\:` should be used to produce literal `:`s.'
   p.option 'root_dir', '--root-dir PATH', String, 'The absolute path to the directory serving your html-files.'
@@ -88,6 +89,7 @@ Mercenary.program(:htmlproofer) do |p|
     options[:validation][:report_mismatched_tags] = opts['report_mismatched_tags'] unless opts['report_mismatched_tags'].nil?
 
     options[:typhoeus] = HTMLProofer::Configuration.parse_json_option('typhoeus_config', opts['typhoeus_config']) unless opts['typhoeus_config'].nil?
+    options[:hydra] = HTMLProofer::Configuration.parse_json_option('hydra_config', opts['hydra_config']) unless opts['hydra_config'].nil?
 
     unless opts['timeframe'].nil?
       options[:cache] ||= {}
--- a/lib/html-proofer/configuration.rb
+++ b/lib/html-proofer/configuration.rb
@@ -78,7 +78,7 @@ module HTMLProofer
       return {} if config.strip.empty?
 
       begin
-        JSON.parse(config)
+        JSON.parse(config, { symbolize_names: true })
       rescue StandardError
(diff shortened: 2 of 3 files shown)


patcon:
Ah, I think I may have found a bug. Seems that when JSON config was being merged from `--typhoeus-config`, it was string keys being merged with symbolized keys internally. Not sure how gracefully this was handled. There were no errors when this happened to typhoeus, but hydra was throwing errors about "max_concurrency" not being valid option.

Screen Shot 2021-03-29 at 5 28 35 PM

(Note the double values at the end of the output debug string in screenshot)

If we instead parse JSON with `JSON.parse(foo, { symbolize_names: true })`, then these merge more cleanly. Going to do that, but lemme know if it should be in a separate PR)

patcon:
Perhaps you could advise on whether this is an acceptable fix, before I start investigating/resolving these test failures. Seems to be an issue with checking a value called "myValue", which may be related to symbolization

gjtorikian:
Ah, thanks for adding this! 

I'm confused why the keys suddenly need to be symbolized, but I don't think that has any effect on the end user, so I'm fine with it. And yes, it looks like the test just needs to be turned into a symbol key rather than a string.

patcon:
yay! thanks! will fix that 🙌 

(Might be a JSON stdlib change way back, maybe just on the nodejs version I'm using)

Should I test a spread of a few ruby versions to be sure? (Not sure if that's done already.)

patcon:
k pushed fix! Should be good to merge. Thanks @gjtorikian 🙌

gjtorikian:
Thanks again!

ArtNow198 (commented):
Iii
Duplicate of #

ArtNow198:
> Powiązane: [# 131](https://github.com/gjtorikian/html-proofer/issues/131)
> 
> Dziękujemy za rozważenie I dziękuję za całą pracę nad utrzymaniem tego narzędzia! :)

joe192:
diff --git a/bin/htmlproofer b/bin/htmlproofer
index 8c4c3362..9d95e4e9 100755
--- a/bin/htmlproofer
+++ b/bin/htmlproofer
@@ -48,6 +48,7 @@ Mercenary.program(:htmlproofer) do |p|
   p.option 'storage_dir', '--storage-dir PATH', String, 'Directory where to store the cache log (default: "tmp/.htmlproofer")'
   p.option 'timeframe', '--timeframe <time>', String, 'A string representing the caching timeframe.'
   p.option 'typhoeus_config', '--typhoeus-config CONFIG', String, 'JSON-formatted string of Typhoeus config. Will override the html-proofer defaults.'
+  p.option 'hydra_config', '--hydra-config CONFIG', String, 'JSON-formatted string of Hydra config. Will override the html-proofer defaults.'
   p.option 'url_ignore', '--url-ignore link1,[link2,...]', Array, 'A comma-separated list of Strings or RegExps containing URLs that are safe to ignore. It affects all HTML attributes. Note that non-HTTP(S) URIs are always ignored'
   p.option 'url_swap', '--url-swap re:string,[re:string,...]', Array, 'A comma-separated list containing key-value pairs of `RegExp => String`. It transforms URLs that match `RegExp` into `String` via `gsub`. The escape sequences `\\:` should be used to produce literal `:`s.'
   p.option 'root_dir', '--root-dir PATH', String, 'The absolute path to the directory serving your html-files.'
@@ -88,6 +89,7 @@ Mercenary.program(:htmlproofer) do |p|
     options[:validation][:report_mismatched_tags] = opts['report_mismatched_tags'] unless opts['report_mismatched_tags'].nil?
 
     options[:typhoeus] = HTMLProofer::Configuration.parse_json_option('typhoeus_config', opts['typhoeus_config']) unless opts['typhoeus_config'].nil?
+    options[:hydra] = HTMLProofer::Configuration.parse_json_option('hydra_config', opts['hydra_config']) unless opts['hydra_config'].nil?
 
     unless opts['timeframe'].nil?
       options[:cache] ||= {}
diff --git a/lib/html-proofer/configuration.rb b/lib/html-proofer/configuration.rb
index efe2f6fc..3a5b3504 100644
--- a/lib/html-proofer/configuration.rb
+++ b/lib/html-proofer/configuration.rb
@@ -78,7 +78,7 @@ def self.parse_json_option(option_name, config)
       return {} if config.strip.empty?
 
       begin
-        JSON.parse(config)
+        JSON.parse(config, { symbolize_names: true })
       rescue StandardError
         raise ArgumentError, "Option '#{option_name} did not contain valid JSON."
       end
diff --git a/spec/html-proofer/parse_json_option_spec.rb b/spec/html-proofer/parse_json_option_spec.rb
index 39075cc9..5228bfbb 100644
--- a/spec/html-proofer/parse_json_option_spec.rb
+++ b/spec/html-proofer/parse_json_option_spec.rb
@@ -37,8 +37,8 @@
 
   it 'Returns an object representing the json when valid json' do
     result = HTMLProofer::Configuration.parse_json_option('testName', '{ "myValue": "hello world!", "numberValue": 123}')
-    expect(result['myValue']).to eq('hello world!')
-    expect(result['numberValue']).to eq(123)
+    expect(result[:myValue]).to eq('hello world!')
+    expect(result[:numberValue]).to eq(123)
   end
 
   it 'Throws an error when the json config is not valid json' do

PR #651, opened 2021-08-29
Add example using proof-html GitHub action

### LINKED ISSUE #642: Add GitHub Action wrappers to README?
Just wanted to share these links in case you weren't aware of them:

- https://github.com/chabad360/htmlproofer
- https://github.com/anishathalye/proof-html

We [used to use](https://github.com/missing-semester/missing-semester/blob/f6c4244bc471bdf9f53b7d0abcbf550dbe310078/.github/workflows/build.yml#L24) a CI setup where we did a `gem install html-proofer` and then called the `htmlproofer` CLI program, but we found it more [convenient](https://github.com/missing-semester/missing-semester/blob/master/.github/workflows/build.yml#L25) to switch to a setup that was a bit more integrated with GitHub actions and let us set parameters that weren't controllable through the CLI program, things like feeding in access tokens, using regex URL ignore patterns, and retrying when checking external links fails.

Sharing this in case you want to add links to the GitHub actions or add our CI setup to the "Real-life examples" section.

anishathalye:
Closes #642


gjtorikian:
Thanks!

Issue #552, opened 2020-02-04
mailto without email shouldn't fail


felicianotech:
I don't believe that HTML Proofer should return an error when a `mailto` URL doesn't include an email address.

Many people including #271 and myself use `mailto` URLs without an email address in order to prep an email to someone, anyone. This is common for a sharing menu where one of the options is to share an article via email.

Latest RFC (I think): https://tools.ietf.org/html/rfc6068

PR #657, opened 2021-10-17
Add flag to allow no-email mailto links

PeterJCLaw:
This defaults to erroring about `mailto` links with email addresses for backwards compatibility, however I'd be equally happy to invert this and have the default allow the no-email case.

Fixes https://github.com/gjtorikian/html-proofer/issues/552.


PeterJCLaw:
@gjtorikian I'm not sure what the lint failure here means (I don't appear to have changed the file it's in); I'm not usually a ruby developer. Could you suggest what I need to do to fix it?

Issue #618, opened 2021-01-10
Add support for site.baseurl, workaround


wgmyers:
When using html-proofer on a Jekyll site with non-empty baseurl such as `/test`, all internal links are reported as broken.

It would be great to be able to specify something like `--internal_baseurl "BASEURL"` on the command line, and then have the checker prepend the value of `BASEURL` to all internal links. This would also locate instances where `BASEURL` was incorrectly not prepended, and be extremely useful to Jekyll users.

Others have found a workaround with the url_swap option, but it does not allow replacing eg `/test' with the empty string or with `/`

```
$ htmlproofer --assume-extension --disable_external --url_swap "/test\:/" ./_site
htmlproofer 3.18.2 | Error:  undefined method `split' for /test\:/:Regexp
```

The workaround working for me is:

```
$ cd _site
$ ln -s . test
$ cd ..
$ htmlproofer --assume-extension --disable_external ./_site
```

kylezervas:
ok thanks im going to try and farm this in to my on

PeterJCLaw:
Not sure if this has since been fixed, but I'm finding that this url swap works for this case: `--url-swap "^/website/:/"`. That's to line up with a Jekyll setting of `baseurl: /website`.

I think the issue in the original example is possibly the use of `\:` rather than `:` to separate the original from the replacement?

The docs on this could definitely be clearer though -- I figured out my url swap by debugging the code.

PR #658, opened 2021-10-18
Document how to adjust for a baseurl

PeterJCLaw:
Fixes https://github.com/gjtorikian/html-proofer/issues/618 (I realise the issue is closed, though hopefully this docs change makes this easier for others in the future)


gjtorikian:
So nice. Thank you!

PR #665, opened 2021-11-28
Run htmlproofer and precede it with the correct ruby binary

dleidert:
We are in the midst of the transition from Ruby 2.7 to 3.0 and run the tests with both versions. But the shebang line in `bin/htmlproofer` will always resolve to the default version, not to the used Ruby version. This small patch makes sure to run it with the Ruby binary belonging to the used environment.


gjtorikian:
Thank you

Issue #581, opened 2020-09-21
Some URLs are reported invalid but are in fact OK


camilleroux:
Here are some URLs that are really working in Firefox but are reported as invalid :

```
 https://doi.org/10.1007/s10072-020-04447-w is an invalid URL
 https://www.researchgate.net/profile/Dan_Ward9/publication/340539075_Sampling_Bias_Explaining_Wide_Variations_in_COVID-19_Case_Fatality_Rates/links/5ea67986299bf11256128426/Sampling-Bias-Explaining-Wide-Variations-in-COVID-19-Case-Fatality-Rates.pdf  is an invalid URL
 https://www.sciencedirect.com/science/article/pii/S1477893920303124 is an invalid URL
 https://doi.org/10.1038/s41586-020-2349-y is an invalid URL
 https://www.sciencedirect.com/science/article/pii/S0001879119301265?casa_token=h2tOm3iqTmYAAAAA:oNK0sZvDVXxE2O1Sr46BiC2zGLnhQ1kQt1DIhX0c-jq6yZsh7fw78cOK0t2zVkqcguscB4BdZ-c is an invalid URL
 https://kjhrm.sljol.info/articles/10.4038/kjhrm.v12i1.41/galley/76/download/ is an invalid URL
 https://doi.org/10.3201/eid2607.200885 is an invalid URL
 http://faculty.washington.edu/jdb/articles/Brown%20(2012,%20PSPB,%20motives).pdf is an invalid URL
 https://www.sciencedirect.com/science/article/pii/S0006291X20314169 is an invalid URL
 https://www.medrxiv.org/content/10.1101/2020.04.07.20045617v1 is an invalid URL
 https://obgyn.onlinelibrary.wiley.com/doi/full/10.1002/ijgo.13182 is an invalid URL
 https://www.who.int/news-room/fact-sheets/detail/the-top-10-causes-of-death  is an invalid URL
 https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7170399/  is an invalid URL
 https://doi.org/10.1016/S1473-3099(20)30287-5 is an invalid URL
 https://www.medrxiv.org/content/10.1101/2020.03.05.20031815v1.full.pdf is an invalid URL
```

mazzystr:
curl shows ` https://doi.org/10.1007/s10072-020-04447-w` to be a redirect

```
$ curl -vvvv  https://doi.org/10.1007/s10072-020-04447-w | more
[...]

PR #666, opened 2021-11-29
Add typhoeus-config user agent and cookies snippets to the README

JackWilb:
I think this might resolve the issues seen in #581 (which is already closed) by allowing cookies in typhoeus.

Adds some snippets to the README to show how to specify a cookie file (from ruby (untested) and the command line) and user agent (from the command line).

I spent some time figuring this out and thought that the information should be in the README for future users. Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,6 +543,28 @@ HTMLProofer.check_directory("out/", {
 }}).run
 ```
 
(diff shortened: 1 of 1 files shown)


gjtorikian:
Cool thank you

Issue #669, opened 2021-12-17
next gen stuff


gjtorikian:
(These are mostly notes for myself.)

html-proofer was one of the first gems I wrote, and...it shows. There are a lot of broken features and confusing hierarchy. I'd like to cut a new release of this gem, ideally within this month, to get rid of or fix the heck out of the following:

- [x] Remove HTML parsing/validation
- [x] Support Ruby 2.7+
- [x] Better hierarchy for internal logic
- [x] Machine parseable error/issue format
- [x] Add some kind of static site mode
- [x] Double check typhoeus_config
- [x] Fix caching to include versioning, in case caching needs to change further
- [x] test extensions
- [x] check commented code / FIXME / TODO
- [x] change attribute_override to swap_attribute
- [x] check_external_hash is now on by default
- [x] check list of links
- [ ] cache keys

Nothing will be made faster, only features that always cause problems need massive reworking.

gjtorikian:
pretty much did it all. still need to write proper release notes

PR #670, opened 2021-12-17
Remove HTML parsing feature

gjtorikian:
What a colossal error this was.

Related to https://github.com/gjtorikian/html-proofer/issues/669.


CarterPape:
RIP HTML checking, the namesake of html-proofer, and the original reason I used this project 🪦

Issue #695, opened 2022-01-09
Problems with internal link cache re-checking logic (4.0.0.rc3)


riccardoporreca:
Cached internal broken links are not re-checked

```bash
rm -rf example_site
mkdir -p example_site
echo '
<!doctype html>
<html>
<head><title>Example</title></head>
<body>
<a href="/missing.html">Missing internal link</a>
</body>
</html>
' > example_site/index.html
```

Running first with empty cache
```bash
rm -rf tmp/.htmlproofer
bundle exec htmlproofer example_site --log-level debug --cache '{"timeframe": "1d"}'
# [...]
# Adding /missing.html to internal cache
# Adding 1 new internal link to the cache
# Removing 0 outdated internal links from the cache
# Checking 1 internal link
# Ran on 1 file!
#
#
# For the Links > Internal check, the following failures were found:
#
# * In example_site/index.html (line 5):
#
#   internally linking to /missing.html, which does not exist
#
#
# HTML-Proofer found 1 failure!
```

Run again: the broken link is not re-checked

```bash
rm -rf tmp/.htmlproofer
bundle exec htmlproofer example_site --log-level debug --cache '{"timeframe": "1d"}'
# Found 1 internal link in the cache
# Adding 0 new internal links to the cache
# Removing 0 outdated internal links from the cache
# Checking 0 internal links
# Ran on 1 file!
#
#
# HTML-Proofer finished successfully.
```

riccardoporreca:
I have been investigating this, which is primarily caused by overriding the metadata, which loses the `found` information (confirmed by `found` being `null` in cache.json after the second round above)

https://github.com/gjtorikian/html-proofer/blob/8107f0cc336d27157f68223ba0467a57d2ce1601/lib/html_proofer/cache.rb#L97-L111

I am trying to propose a fix following my investigations

riccardoporreca:
See PR #696 

Note that there is still one main flaw even after the proposed fix, which re-checks an internal url if not all cached checks were OK:

https://github.com/gjtorikian/html-proofer/blob/287bd2d52c69b23ce579eeb8e13b1a12a1103f46/lib/html_proofer/cache.rb#L104

This is not enough, since a relative URL that was OK for all the files previously checked and cached does not provide any guarantee against e.g. the same URL being OK for a new file (for instance, this could happen if a file is moved and the relative urls in it would not be updated). This is very related to https://github.com/gjtorikian/html-proofer/issues/671#issuecomment-1003456160

[...]

PR #696, opened 2022-01-09
Fix internal cache metadata check

riccardoporreca:
Fixes #695

* Unit test revealing issue #695
* Fix logic for checking cached failures and managing new url metadata

PR #694, opened 2022-01-08
Feature/fix files sources processing

riccardoporreca:
Fixes #693 

- Fix merging urls across checks by concatenating their metadata
- Preserve and use file-specific source information

PR #697, opened 2022-01-09
Restore pr 696

gjtorikian:
A resurrection of the commits previously in https://github.com/gjtorikian/html-proofer/pull/696, which I accidentally forced push to, because I was looking at #694 at the same time in my tabs. 🤦 

Should close https://github.com/gjtorikian/html-proofer/issues/695


gjtorikian:
@riccardoporreca Were the tests broken before, I did I break them?

lib/html_proofer/cache.rb
```diff
@@ -194,7 +194,7 @@ def size(type)
                    elsif cache_version != CACHE_VERSION
                    # if cache version is newer...do something
                    else
-                     log[:internal] = log[:internal].transform_keys { |key| JSON.parse(key.to_s, symbolize_names: true) }
```
riccardoporreca: @gjtorikian, We should still need to transform the JSON-string key back to a hash as in `internal_urls`. The initial `.to_s` however would not be needed.

gjtorikian:
Been fighting with this for a few days now, I think I see a path forward but untangling it from this code I messed up is proving to be a challenge. Closing this and will open up a replacement PR soon.

PR #711, opened 2022-06-07
Use <path>:<line> in the CLI reporter

riccardoporreca:
* This is recognized by many terminals / IDEs, and allows reaching the line in the file from the reported error message.


riccardoporreca:
@gjtorikian, this is another small suggestion was thinking of in the past, that would improve the experience with reported errors, perhaps something you would consider for the new release.

PR #726, opened 2022-07-13
Change --url-ignore to --ignore-urls in README.md

nwhetsell:
This PR updates the README to use `--ignore-urls` instead of `--url-ignore` for compatibility with HTMLProofer v4.


gjtorikian:
Thanks!

Issue #733, opened 2022-07-20
Cached external URLs with missing hash are not re-checked


riccardoporreca:
How to reproduce based on this minimal `example.html`
```html
<html>
<head><title>Example</title></head>
<body>
<a href="https://github.com/gjtorikian/html-proofer#readme">Existing external hash</a>
<a href="https://github.com/gjtorikian/html-proofer#missing-hash">Non-existing external hash</a>
<div id="existing-hash"></div>
<a href="#existing-hash">Existing internal hash</a>
<a href="#missing-hash">Non-existing internal hash</a>
</body>
</html>
```
First run populating the cache
```sh
rm -rf tmp/.htmlproofer/
bundle exec htmlproofer example.html --cache '{ "timeframe": { "external": "1d", "internal": "1d"}}'
# For the Links > External check, the following failures were found:
# * At example.html:5:
#   External link https://github.com/gjtorikian/html-proofer#missing-hash failed: https://github.com/gjtorikian/html-proofer exists, but the hash 'missing-hash' does not (status code 200)
# For the Links > Internal check, the following failures were found:
# * At example.html:8:
#   internally linking to #missing-hash; the file exists, but the hash 'missing-hash' does not
# HTML-Proofer found 2 failures!
```
2nd run, failures should be rechecked, but it is not the case for the external link:
```sh
bundle exec htmlproofer example.html --cache '{ "timeframe": { "external": "1d", "internal": "1d"}}'
# For the Links > Internal check, the following failures were found:
# * At _examples/example.html:8:
#   internally linking to #missing-hash; the file exists, but the hash 'missing-hash' does not
# HTML-Proofer found 1 failure!
```

The cache content reveals that despite storing the error, `found` is `true` and as such not detected for re-checking:
```json
{
    "version": 2,
    "internal": {
        "#existing-hash": {
            "time": "2022-07-20 20:01:53 +0200",
            "metadata": [
                {
                    "source": "_examples/example.html",
                    "filename": "_examples/example.html",
                    "line": 7,
                    "base_url": "",
                    "found": true
                }
            ]
        },
        "#missing-hash": {
            "time": "2022-07-20 20:01:53 +0200",
            "metadata": [
                {
                    "source": "_examples/example.html",
                    "filename": "_examples/example.html",
                    "line": 8,
                    "base_url": "",
                    "found": false
                },
                {
                    "source": "_examples/example.html",
                    "filename": "_examples/example.html",
                    "line": 8,
[...]

PR #736, opened 2022-07-23
Cache fixes and improvements

### LINKED ISSUE #737: Disable internal/external cache altogether if not timeframe is specified
This follows up on #708, where support for separate timeframe for internal/external cache was introduced.

At the moment, if any of the two is not specified, the corresponding cache is still enabled but considered as always expired. Instead, one can disable the corresponding cache.

riccardoporreca:
While trying to look at the issue spotted in #733

>  the rechecked internal link seems to duplicate the metadata (every run would include new metadata

I took the chance to review the current cache deletion / addition logic also following up on some of the review comments from #712 that affect the (mis)behaviour beyond performance considerations.

This also includes a small extension that would close #737.

I tried to be very detailed with individual commits and verbose messages describing the reason for each change, so this PR is best reviewed by looking at diffs and messages from individual commits.


gjtorikian (approved):

gjtorikian:
Looks good, thank you!

Issue #724, opened 2022-07-12
srcset width/pixel density descriptors cause missing image error


fallax:
The [spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source#attr-srcset) for the srcset attribute says:

> A list of one or more strings, separated by commas, indicating a set of possible images represented by the source for the browser to use. Each string is composed of:
> 
> -    One URL specifying an image.
> -    A width descriptor, which consists of a string containing a positive integer directly followed by "w", such as 300w. The default value, if missing, is the infinity.
> -    A pixel density descriptor, that is a positive floating number directly followed by "x". The default value, if missing, is 1x.
> 
> Each string in the list must have at least a width descriptor or a pixel density descriptor to be valid.

However when a width descriptor or pixel density descriptor is used, for example:

```
<html>
<body>
    <img
        srcset="test.webp 1.5x"
        alt="test"
        width="100"
        height="150"
        />
</body>
</html>
```

html-proofer says ```internal image test.webp 1.5x does not exist``` even if test.webp is present.

Additionally, most tests for srcsets (e.g. /spec/html-proofer/fixtures/images/multiple_srcset_success.html) do not include either a width descriptor or pixel density descriptor, which the spec for the attribute says it must have. I'm guessing this is why this issue has crept in?

gjtorikian:
Whoops, thanks for the excellent report and repro case. 4.1 should fix this and adds some more checks for more legitimate `srcset` attributes.

fallax:
Thanks for the quick fix - this has allowed us to upgrade :)

PR #744, opened 2022-07-29
Srcsets - better handling of multiple srcsets

fallax:
The previous fix to #724 didn't handle this situation:

`srcset="test.webp 1.5x, test2.webp 2x"`

This PR should fix this, add a new test case, and hopefully remove some code duplication as a nice bonus.

### DIFF
--- a/lib/html_proofer/check/images.rb
+++ b/lib/html_proofer/check/images.rb
@@ -25,7 +25,7 @@ module HTMLProofer
               content: @img.content)
           elsif @img.multiple_srcsets? || @img.multiple_sizes?
             @img.srcsets_wo_sizes.each do |srcset|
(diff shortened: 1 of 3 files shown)


gjtorikian:
Oh, thanks very much! Releasing this as a patch update right now.

fallax:
Looks like this still leaves a few issues.  tags with multiple pixel density srcset attributes now work correctly - but <source> tags don't.

I think this can be fixed by switching <source> tags to being handled by the code in images.rb rather than the code in links.rb. I'm happy to contribute a PR to do this. Does that sound like a sensible approach to you? If so I'll get a second PR going...

gjtorikian:
Sure, I can review the change and see what makes most sense then.

PR #747, opened 2022-08-01
Switch 'source' elements to use image check code path

fallax:
As discussed in comments to #744 - this change shifts 'source' elements to be handled by the code path for checking images.

A number of other small changes to tests needed to be made also - this was because these tests were referring to non-existent images in tests that did not relate to the URL not being to a valid image. These have been updated to point to images that actually exist.


gjtorikian:
Thank you, again.

Issue #754, opened 2022-08-13
Mailto check failed in some cases


Atarity:
Source: `<a href="mailto:mike@snnkv.com">mike@snnkv.com</a>`

Result: 
```
[14:40:12] htmlProofer: * At _site/now/index.html:87:
[14:40:12] htmlProofer:
[14:40:12] htmlProofer:   mailto:mike@snnkv.com contains an invalid email address
```

While I change domain name a bit (eg `@snkv.com` or `@snnnkv.com`) it pass the test. The problem is: it is impossible to just turn such check off, so htmlproofer became useless for me since v4.

riccardoporreca:
@Atarity, I cannot reproduce with v4.4.0 using the following `mailto.html`

```html
<!DOCTYPE html>
<html>
  <body>
    <a href="mailto:mike@snnkv.com">mike@snnkv.com</a>
    <a href="mailto:mike@snkv.com">mike@snkv.com</a>
    <a href="mailto:mike@snnnkv.com">mike@snnnkv.com</a>
  </body>
</html>
```

```sh
bundle exec htmlproofer mailto.html
# HTML-Proofer finished successfully
```

Are you perhaps using other options that might interfere with the check?

> The problem is: it is impossible to just turn such check off

You can do so via `--ignore-urls "/^mailto:/"`

Hope this helps

Atarity:
Thanks for `--ignore-urls` workaround. Yes, I actually use lots of other options:
```
--extension .html
--assume-extension .html
--checks "Links,Scripts,OpenGraph"
--enforce-https=false
--ignore-status-codes "999,401"
--disable-external=true
--ignore-urls "/^mailto:/"
```

riccardoporreca:
None of the options seem to interfere with the minimal example above... even with those flags the issue cannot be reproduced.
Hard to investigate if not reproducible, you can perhaps try updating all gems.

Atarity:
I performed one more test. Took basic `mailto.html` you provided above. Process the file with `htmlproofer` as you do. Success. But as soon as I check this file as a part of my website it is the same result. It is failed in only one email address out of 3. 
```
[10:53:44] htmlProofer: For the Links check, the following failures were found:
[10:53:44] htmlProofer:
[10:53:44] htmlProofer: * At _site/mailto.html:4:
[10:53:44] htmlProofer:
[10:53:44] htmlProofer:   mailto:mike@snnkv.com contains an invalid email address
```
[...]

PR #755, opened 2022-08-14
Fix and improve swap_attribute README example

riccardoporreca:
* Wrongly specified as `swap_url`
* Using `data-src` as a more meaningful case than `srcset`, as the latter is in any case allowed instead of `src`

See also https://github.com/gjtorikian/html-proofer/issues/754#issuecomment-1214425422


gjtorikian:
Thank you!

PR #760, opened 2022-09-14
Lock to >= Ruby 3.1

gjtorikian:
Ruby 3.1 has some optimizations for Windows; since I'm breaking a new major release out, it makes sense to lift the restrictions here a bit. Plus, I want to use `async` instead of `parallel`, which is a Ruby 3+ addition.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,10 @@
 name: Ruby CI
 
 on:
+  pull_request:
   push:
+    branches:
+      - main
 
 permissions:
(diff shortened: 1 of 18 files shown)


SeanKilleen:
@gjtorikian I support doing this, but FYI I had to lock html-proofer to v4 in order to keep my Jekyll site working with GitHub pages (which is pegged to ruby 2.7.4 at the moment). 

Might be worth adding a note on that to the release notes, given the possible size of the audience that would experience it; your call.

PR #767, opened 2022-09-22
Validate options

asbjornu:
If you, for instance, pass `assume_extension: true` to HTMLProofer now, it is going to fail with the following error:

```ruby
NoMethodError: undefined method `empty?' for true:TrueClass
    ./lib/html_proofer/attribute/url.rb:138:in `file_path'
    ./lib/html_proofer/utils.rb:12:in `blank?'
    ./lib/html_proofer/attribute/url.rb:128:in `absolute_path'
    ./lib/html_proofer/check/links.rb:57:in `block in run'
```

This error is a bit hard to decipher since it is not pointed to the fact that the problem stems from an invalid type being passed in `options` and it also doesn't say which option key is problematic. This PR adds validation of options on initialization so more user-friendly and descriptive errors are thrown when options have invalid values.


lib/html_proofer/configuration.rb
```diff
@@ -86,6 +88,24 @@ def parse_json_option(option_name, config, symbolize_names: true)
           raise ArgumentError, "Option '#{option_name} did not contain valid JSON."
         end
       end
+
+      private
+
+      def default_options
+        PROOFER_DEFAULTS.merge(typhoeus: TYPHOEUS_DEFAULTS).merge(hydra: HYDRA_DEFAULTS).merge(parallel: PARALLEL_DEFAULTS)
```
asbjornu, line 95: This does not feel like an elegant or bullet-proof way to enumerate all existing options to extract their default value type. Suggestions for how to improve this would be appreciated.

asbjornu (commented):
I can see that I've managed to add the file `spec/html-proofer/fixtures/links/erstiebegrüßung.html`. I have major issues with that file on macOS 11.7. I've attempted to do [everything mentioned in this post](https://www.git-tower.com/help/guides/faq-and-tips/faq/unicode-filenames/mac), but the file is still deleted and re-added at random and makes other Git operations such as rebase, impossible.

asbjornu:
As #772 is merged, this PR has now been rebased on `HEAD` of `main`, the file `erstiebegrüßung.html` is gone, and the PR should be ready for review.

gjtorikian:
Thanks! I'm going to release this as a 4.x bug fix for this and then work more on V5 (https://github.com/gjtorikian/html-proofer/pull/759) this weekend.

lib/html_proofer.rb
```diff
@@ -25,22 +25,22 @@ def check_file(file, options = {})
       raise ArgumentError unless file.is_a?(String)
       raise ArgumentError, "#{file} does not exist" unless File.exist?(file)
 
-      options[:type] = :file
```
Mihail211292, line 28: options.is_a?(Hash)+ ({yes}))

Issue #768, opened 2022-09-22
HTMLProofer times out


asbjornu:
After upgrading to HTMLProofer v4, it has stopped working for [our Jekyll-built site](https://developer.swedbankpay.com/). As can be seen in [the following build](https://github.com/SwedbankPay/developer.swedbankpay.com/actions/runs/3101420585/jobs/5022794462#step:6:4446), the build step is canceled before completion, after having run for 6 hours. It halts after the following lines have been logged:

```
ETHON: performed MULTI
Checking 3341 internal links
```

I would have run HTMLProofer on the command line if `before_request` was available there, but as that requires HTMLProofer to be run from Ruby, I've written [the following little wrapper class](https://github.com/SwedbankPay/developer.swedbankpay.com/blob/94adc65cfba2d91043a3dc6fea7d60037d62067a/spec/support/verifier.rb) (abbreviated here for brevity):

```ruby
class Verifier
  def verify(path)
    proofer = HTMLProofer.check_directory(path, options)
    proofer.before_request { |request| before_request(request) }
    proofer.run
  end

  private

  def before_request(request)
    uri = URI(request.base_url)

    return unless uri.host.match('github\.(com|io)$')

    auth = "Bearer #{@auth_token}"
    request.options[:headers]['Authorization'] = auth
  end

  def options
    {
      cache: { timeframe: { external: '1w' } },
      check_html: true,
      check_unrendered_link: true,
      checks: %w[Links Images Scripts UnrenderedLink],
      enforce_https: true,
      log_level: @log_level,
      only_4xx: true,
      parallel: { in_processes: Concurrent.processor_count },
      url_ignore: [
        'https://api.payex.com/',
        'http://www.wikipedia.org',
        'http://restcookbook.com/Basics/loggingin/'
      ]
    }
  end
end
```

If you can spot anything erroneous about the above code, I would highly appreciate any pointers. To reproduce the problem, clone [developer.swedbankpay.com](https://github.com/SwedbankPay/developer.swedbankpay.com) and then run:

```shell
# First build the site with Jekyll. Use Docker so Java dependencies, etc., don't have to be installed locally.
# Ignore the version of HTMLProofer being installed here, it's inside the Docker container and unrelated
# to the tests executed with Rake.
docker compose run portal build --site-url=https://developer.swedbankpay.com 
bundle exec rake
```

If there's anything I can do to help debug this issue, please let me know!

riccardoporreca:
[...]

PR #770, opened 2022-09-23
Optimize checking internal link hashes in target files

riccardoporreca:
By collecting non-duplicate hashes by file and calling the time-consuming `create_nokogiri()` only once per file.

Should address #768

See the attached output ([htmlproofer-output.txt](https://github.com/gjtorikian/html-proofer/files/9637319/htmlproofer-output.txt)) of running the internal link checks on the `build-site` site artifact discussed in https://github.com/gjtorikian/html-proofer/issues/768#issuecomment-1255628337, via
```sh
htmlproofer --disable-external true --log-level debug --checks Links  build-site/
```

- The total execution is now around 10 min
- The output shows also the additional info / debug logs that help tracking how much htmlproofer is doing for checking internal links / hashes
  - in particular, it reflects the new approach of checking internal hashes by target file
  - more in general, it reports at debug level how many individual references to the same internal URL ("metadata") we iterate we have to iterate over

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -24,7 +24,7 @@ task :proof_readme do
   require "redcarpet"
 
   renderer = Redcarpet::Render::HTML.new(\
-    with_toc_data: true
+    with_toc_data: true,
   )
   redcarpet = Redcarpet::Markdown.new(renderer)
   html = redcarpet.render(File.read("README.md"))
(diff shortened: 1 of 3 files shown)


gjtorikian:
Wow. Beautiful. Thank you so very much for this work~!

riccardoporreca:
My pleasure @gjtorikian!

Issue #783, opened 2022-12-02
Buggy parsing of srcset attribute


anishathalye:
The implementation currently [splits the `srcset` on "," (comma)](https://github.com/gjtorikian/html-proofer/blob/7648d24c61099ae598235250b1a9a57ad1d2eed0/lib/html_proofer/element.rb#L90), but this is not quite correct.

This results in incorrect handling of URLs with commas in the `srcset`. For example, base64-encoded images will be incorrectly handled, because it'll split on the comma in e.g., `data:image/png;base64,iKB...=`, and then the later check will say that the base64-encoded data `iKB...=` on its own is not a valid URL (that second part won't be caught by [`base64?`](https://github.com/gjtorikian/html-proofer/blob/bea6b3e24593add1345587fcc8c44fa2ad2a6028/lib/html_proofer/attribute/url.rb#L123) because the first part has been split off).

It'll also fail with regular old URLs that contain commas.

[This page](https://html.spec.whatwg.org/#parsing-a-srcset-attribute) describes how the srcset should be parsed.

gjtorikian:
Thanks forthe report, a new patch release will go out to address this.

PR #784, opened 2022-12-05
Fix poor srcset parsing

gjtorikian:
Closes https://github.com/gjtorikian/html-proofer/issues/783


anishathalye:
That was fast, thank you for the quick fix!

Issue #787, opened 2023-01-08
Crashing when checking the external hashes of URLs 


stevecheckoway:
In order to check if a URL's fragment is valid, html-proofer uses Nokogiri to parse the document at the URL. Nokogiri has a [default maximum tree depth of 400](https://github.com/sparklemotion/nokogiri/blob/main/lib/nokogiri/gumbo.rb#L13). (This can be [changed](https://nokogiri.org/rdoc/Nokogiri/HTML5.html#module-Nokogiri::HTML5-label-Maximum+tree+depth).)

Since the content of external documents is outside the control of the user of html-proofer, it makes sense to catch any exceptions that result from parsing those (and probably all) documents with Nokogiri.

~I don't have a URL off-hand~ that demonstrates this as it is one of 900 on my website. But I did confirm that `htmlproofer --check-external-hash` fails with the error
```
/path/vendor/bundle/ruby/3.1.0/gems/nokogiri-1.13.10-arm64-darwin/lib/nokogiri/html5/document.rb:85:in `parse': Document tree depth limit exceeded (ArgumentError)
        from /path/vendor/bundle/ruby/3.1.0/gems/nokogiri-1.13.10-arm64-darwin/lib/nokogiri/html5/document.rb:85:in `do_parse'
        from /path/vendor/bundle/ruby/3.1.0/gems/nokogiri-1.13.10-arm64-darwin/lib/nokogiri/html5/document.rb:43:in `parse'
        from /path/vendor/bundle/ruby/3.1.0/gems/nokogiri-1.13.10-arm64-darwin/lib/nokogiri/html5.rb:31:in `HTML5'
        from /path/vendor/bundle/ruby/3.1.0/gems/html-proofer-5.0.3/lib/html_proofer/utils.rb:22:in `create_nokogiri'
...
```
and `htmlproofer --no-check-external-hash` completes successfully.

**Update:** I realized I could change the log level to figure out which URL was causing the problem. In fact, the issue was a link to a PDF with a query fragment, similar to #663.

To reproduce put
```html
<!DOCTYPE html>
<a href='https://checkoway.net/teaching/cs210/2021-fall/slides/Lec19%20ClocksandTiming.pdf#page=25'>X</a>
```
in `a.html` and run `$ htmlproofer --log-level :debug a.html`.

Trimmed output:
```
Running 3 checks (Images, Links, Scripts) in a.html on *.html files ...


Running Images in a.html
Running Links in a.html
Running Scripts in a.html
[...]

PR #789, opened 2023-01-20
Improved PDF hash handling

gjtorikian:
Closes https://github.com/gjtorikian/html-proofer/issues/787


stevecheckoway:
Would you mind changing the tests so that they're not fetching PDFs from my website? I provided that just as an example of a link that failed because the PDF was parsed as html. I'm definitely not committing to maintaining slides for old classes I've taught at the same URL forever.

gjtorikian:
The test suite uses VCR to record the (expensive) HTTP call once, and then simply reuses that fixture any time the test is run again: https://github.com/gjtorikian/html-proofer/blob/main/spec/html-proofer/fixtures/vcr_cassettes/links/pdf_w_hash_broken_html_log_level_error_type_file_.yml

Is that ok?

stevecheckoway:
That contains the complete PDF. I don't own the contents of that PDF (they are lecture slides I use with permission from the author). Please use a different PDF.

gjtorikian:
Ok in the future please provide a test case that is safe to reproduce please.

Issue #792, opened 2023-02-25
Reporting broken links when using relative internal links reported as broken


bnorberg:
Hi, I am running into an issue where HTML Proofer is reporting working relative links as broken.

### Example link
```<a href="../admin/configuring-load-balancer-endpoints.html">```

### Example feedback from HTML Proofer
![Screenshot 2023-02-25 at 11 05 53 AM](https://user-images.githubusercontent.com/367573/221367176-c412ea5d-b767-437f-bdad-0a08435db248.png)

Any suggestions on what could be causing the broken link message? 

My only solution to ensure I don't get false positives on relative links is to add '../' links in the ignore-urls option or changing all links to root relative. Neither solution is ideal.

gjtorikian:
Hm, I can't reproduce this issue. Are you sure `configuring-load-balancer-endpoints.html` exists? Could you zip up your output for me to examine it, or is there some other way I can reproduce it?

bnorberg:
Thanks, for the quick response, [gjtorikian](https://github.com/gjtorikian).

That page does exist because the link works on the deploy site. Here's the page in the gh-pages branch of the deploy, https://github.com/NetAppDocs/storagegrid-116/blob/gh-pages/admin/configuring-load-balancer-endpoints.html and here is the page that is reference the above link through the relative https://github.com/NetAppDocs/storagegrid-116/blob/gh-pages/admin/configuring-custom-server-certificate-for-storage-node-or-clb.html. Again this is just one example. We are getting the broken link on all `..` relative links. If we change `<a href="../admin/configuring-load-balancer-endpoints.html">` to `<a href="./admin/configuring-load-balancer-endpoints.html">` or `<a href="/admin/configuring-load-balancer-endpoints.html">`, the link checker will not return a broken link message.

We are using a mix of private and public GitHub build runners and different deployment environments for site builds so it will probably not be possible to replicate what I am seeing.

Is there a possibility that this could be related to a site prefix causing the `..` in the relative link to not actually be in the correct directory to find the linked file?? We use a language prefix on our [site](https://docs.netapp.com/us-en/storagegrid-116/admin/configuring-load-balancer-endpoints.html).

Thanks again.

gjtorikian:
Sorry, but this is probably going to be a bit of back and forth. 😓 

[...]

PR #793, opened 2023-03-06
Fix cleaning relative links cache key

riccardoporreca:
See https://github.com/gjtorikian/html-proofer/issues/792#issuecomment-1457199390


lib/html_proofer/cache.rb
```diff
@@ -93,8 +93,10 @@ def retrieve_urls(urls_detected, type)
       # if there are no urls, bail
       return {} if urls_detected.empty?
 
-      urls_detected = urls_detected.transform_keys do |url|
-        cleaned_url(url)
+      if type == :external
+        urls_detected = urls_detected.transform_keys do |url|
+          cleaned_url(url)
+        end
       end
```
riccardoporreca, line 100: Actually, I think we may want to only avoid `escape_unescape` for the external links, since the other part of `cleaned_url` that handles trailing "/", "#", "?" would still be relevant for internal links. Any input on this @gjtorikian?
If we keep part of `cleaned_url` for internal links, we should probably also include it to `add_internal` for consistency, similar to what is done for `add_external`

riccardoporreca, line 100: Actually, I think we may want to only avoid `escape_unescape` for the external links, since the other part of `cleaned_url` that handles trailing "/", "#", "?" would still be relevant for internal links. Any input on this @gjtorikian?
If we keep part of `cleaned_url` for internal links, we should probably also include it to `add_internal` for consistency, similar to what is done for `add_external`

Issue #757, opened 2022-09-08
Anchor links are encoded in html-proofer output


mattcone:
Hey there! In our translated files, Japanese anchor links are being encoded in the html-proofer output even though the anchor links are not encoded in our HTML. So, for example, we have the following in an HTML file:

```html
<a href="#authorization-ヘッダーを作成する">Authorization ヘッダーを作成する</a>
```

But see the following in the html-proofer output:

```bash
internally linking to #authorization-%E3%83%98%E3%83%83%E3%83%80%E3%83%BC%E3%82%92%E4%BD%9C%E6%88%90%E3%81%99%E3%82%8B; the file exists, but the hash ‘authorization-%E3%83%98%E3%83%83%E3%83%80%E3%83%BC%E3%82%92%E4%BD%9C%E6%88%90%E3%81%99%E3%82%8B’ does not
```

I'm not sure if this is a bug with html-proofer or a configuration problem on our end, but I don't recall seeing this before. Thanks for any help you can provide.

gjtorikian:
Oh I’m sorry I missed this. Was the problem fixed?

ahpook:
I am seeing this same problem w/ html-proofer 5.0.8

```
* At _site/de/best-practices/index.html:591:

  internally linking to building-community/#teilen-sie-die-eigent%C3%BCmerschaft-an-ihrem-projekt, which does not exist

```

I note the encoding is correct in the debug output as it's checking:

```
(354 / 1737) Internal link ../building-community/#teilen-sie-die-eigentümerschaft-an-ihrem-projekt: Checking 2 references
```

ahpook:
I have done some more investigation on this issue. I think it's caused by an upstream dependency rather than something in html-proofer itself.

You can see the error in action on this test run : https://github.com/github/opensource.guide/actions/runs/8887912548/job/24403938507?pr=3214 

I added one of the failing links from that project as a test fixture, and added a test to exercise it, which you can see on my fork:
https://github.com/ahpook/html-proofer/commit/7da8d0436b940c1a2b203ae81889e85df553131b

That test passes on my branch. And I'm using the same ruby version on both html-proofer and opensource.guide. Looking through the Gemfile, one _possible_ difference is that html-proofer is using `unicode-display_width (2.5.0)` and opensource.guide is at `unicode-display_width (1.8.0)`, I think due to a jekyll dependency - I'm not familiar with this gem but was just looking for anything i18n/encoding related that could explain the discrepancy.

[...]

PR #842, opened 2024-05-23
Check utf8 relative anchors

ahpook:
This adds a test for the false-negative failures seen in https://github.com/gjtorikian/html-proofer/issues/757 - the root cause was an interaction between the (now-removed) regexp in the `clean_url!` method and Addressable::URI's `normalize`. The test passes now that https://github.com/gjtorikian/html-proofer/pull/820 is merged but the test should prevent regressions in the future.


gjtorikian:
Rad, thank you!

PR #875, opened 2026-07-06
Create SECURITY.md

tejedaarscelis44:


gjtorikian:
thanks, but there's not much here I would accept.

